PR #28, opened 2023-06-22
@W-13052381 - now able to package and install in vscode running instance.

dbreese:
`vsce package` is the command that creates the package which automatically runs the `vscode:prepublish` task defined in package.json. Currently I have this set up to successfully compile, lint (for errors), and test before packaging occurs.

Next steps: bundle everything into a single file for performance as suggested by VSCode packager: 

```
This extension consists of 14496 files, out of which 12975 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension
```

@khawkins @sfdctaka @maliroteh-sf


.vscodeignore
```diff
@@ -1,6 +1,6 @@
 .vscode/**
 .vscode-test/**
-src/**
+src/**/*.ts
```
dbreese, line 3: Otherwise, the *.html and *.js files are not copied from instructions dir.

package.json
```diff
@@ -58,6 +72,7 @@
     "@typescript-eslint/parser": "^5.56.0",
     "@vscode/l10n-dev": "^0.0.26",
     "@vscode/test-electron": "^2.3.0",
+    "esbuild": "^0.18.6",
```
dbreese, line 75: Opps, this is not needed just yet as this is what does the bundling. Will remove until next PR.

package.json
```diff
@@ -1,15 +1,28 @@
...
   ],
+  "qna": "https://github.com/salesforce/salesforce-offline-vscode/issues",
+  "bugs": {
+    "url": "https://github.com/salesforce/salesforce-offline-vscode/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/salesforce/salesforce-offline-vscode.git"
+  },
+  "icon": "images/salesforce.png",
```
maliroteh-sf, line 21: if you rename the file to something like `pkg-icon.png` then in the future if the icon changes we can just copy it over and submit a PR without having to rename it here too.

package.json
```diff
@@ -36,15 +50,15 @@
...
     "compile": "tsc -p ./",
     "watch": "tsc -watch -p ./",
     "pretest": "npm run compile && npm run lint",
     "lint": "eslint src --ext ts",
     "test": "node ./out/test/runTest.js",
     "test-coverage": "node ./out/test/runTest.js --coverage",
     "prettier:write": "prettier --write \"src/**/*.{ts, js}\"",
-    "prettier:verify": "prettier --list-different \"src/**/*.{ts, js}\""
+    "prettier:verify": "prettier --list-different \"src/**/*.{ts, js}\"",
+    "vscode:prepublish": "npm run pretest && npm run test && npm run compile"
```
maliroteh-sf, line 61: `pretest` already does the compilation so no need to to do another compilation at the end

maliroteh-sf (approved):

sfdctaka (approved):
LGTM!

package.json
dbreese: Good idea, will update!

package.json
dbreese: No sure why I saw lint run twice though. I must be hallucinating. Still a good change though!

package.json
```diff
@@ -1,15 +1,28 @@
 {
   "name": "salesforcedx-vscode-mobile",
-  "displayName": "Salesforce Mobile Extensions for Visual Studio Code",
+  "displayName": "%extension.displayName%",
   "description": "",
   "version": "0.0.1",
-  "publisher": "salesforce",
+  "publisher": "Salesforce",
```
khawkins, line 6: This was correct the original way. Cf. https://github.com/forcedotcom/salesforcedx-vscode/blob/c8c2b544703d8b1b2f5dc072d6c289aa3e6c51dd/packages/salesforcedx-vscode-core/package.json#L19

package.json
```diff
@@ -18,6 +31,7 @@
   "files": [
     "/out",
     "./l10n",
+    "./images",
```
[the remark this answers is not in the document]
dbreese: I agree. I can move the html. and .js files to something like `/static`? as part of this PR, or can do it as a follow up if it might step on things others are already doing?

package.json
dbreese: If I dont change it, it shows up lower case in the extensions, which is why I changed it, but maybe there is another way. Will investigate:

image

khawkins: You can see in the URL (e.g. https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode) that the name is lowercase, so there must be something else that's governing the display of that name.

package.json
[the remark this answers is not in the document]
khawkins: I'm okay either way. I think it's not fluctuating at this point, so you're welcome to do it here if you want. If it would be better focused as its own thing, that's okay too.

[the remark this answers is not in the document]
dbreese: I'd rather do in another PR. Having a lot of questions/issues trying to get esbuild bundles to work, so will focus on that first.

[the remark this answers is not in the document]
khawkins: Makes sense to me! LMK if I can help with esbuild issues. I don't know anything, but happy to be a set of 👀 if it's helpful.

khawkins (approved):
Looks good so far! I'm sure the fun part is yet to come. 🤭

PR #29, opened 2023-06-23
Content updates, bug fix

khawkins:
- Updating various content sets per CX updates.
- Fixed a bug in the workflow where, if the user selects an existing project folder that's already loaded into the workspace, the workflow won't continue because it's expecting the workspace to reload—which it won't in this case.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -20,50 +20,73 @@ const onboardingWizardStateKey =
     'salesforcedx-vscode-offline-app.onboardingWizard.projectCreationState';
 
 enum OnboardingWizardState {
-    projectCreated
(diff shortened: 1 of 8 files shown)


src/extension.ts
```diff
@@ -20,50 +20,73 @@ const onboardingWizardStateKey =
     'salesforcedx-vscode-offline-app.onboardingWizard.projectCreationState';
 
 enum OnboardingWizardState {
-    projectCreated
+    projectConfigured
```
khawkins, line 23: This isn't a state strictly associated with project creation anymore.

sfdctaka (approved):
LGTM!

PR #32, opened 2023-06-29
GHA to create a release

khawkins:
First step to publishing! This creates a GitHub Action that will be manually triggered from the Actions tab. When run, this action will create a GitHub release on the repo, based on the package version in `package.json`. It will then build the VSIX package, and attach that package to the release as an artifact.

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -1,0 +1,58 @@
+name: Create GitHub Release for Extension Package
+
+on:
+    workflow_dispatch:
+
+jobs:
+    release:
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v3
+
+            - name: Install Node.js
+              uses: actions/setup-node@v3
+              with:
+                  node-version: '16.x'
+
+            - name: Install dependencies
+              run: npm install
+
+            - name: Build VSIX Package
+              run: npx vsce package
+
+            - name: Get package version
+              id: get_package_version
+              run: echo "package_version=$(node -p "require('./package.json').version")" >> "$GITHUB_OUTPUT"
+
+            - name: Create release
+              id: create_release
+              uses: actions/github-script@v6
+              with:
+                  script: |
+                      const fs = require('fs');
+                      const response = await github.rest.repos.createRelease({
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
+                        tag_name: 'v${{ steps.get_package_version.outputs.package_version }}',
+                        name: 'Release ${{ steps.get_package_version.outputs.package_version }}',
+                        body: 'Generating release for version ${{ steps.get_package_version.outputs.package_version }}'
+                      });
+                      return response.data.id;
+            - name: Attach VSIX package to release
+              uses: actions/github-script@v6
+              with:
+                  script: |
+                      const fs = require('fs');
+                      const packageJson = require('./package.json');
+                      const packageName = packageJson.name;
+                      const packageVersion = packageJson.version;
+                      const vsixFilename = `${packageName}-${packageVersion}.vsix`;
+                      const uploadUrl = await github.rest.repos.uploadReleaseAsset({
+                        owner: context.repo.owner,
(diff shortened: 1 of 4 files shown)


.github/workflows/create_release.yml
```diff
@@ -0,0 +1,58 @@
+name: Create GitHub Release for Extension Package
+
+on:
+    workflow_dispatch:
```
khawkins, line 4: One thing we could do here, which is currently hard-coded [below](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/32/files#diff-20b4af9a743edb3be348e23e792bf3fb0ad92637b44fafa6c5de878558abce7dR40), is give an input field to the workflow, that would allow the initiating user to create the description associated with the GitHub Release. Could be a nice way to make the "release notes" applicable to each release. Easy to do.

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,58 @@
...
+              with:
+                  node-version: '16.x'
+
+            - name: Install dependencies
+              run: npm install
+
+            - name: Build VSIX Package
+              run: npx vsce package
+
+            - name: Get package version
```
khawkins, line 25: Realizing after doing more of this script, that I could just inline this specific step to the "Create release" step below. I'll make that update, as I think that will be cleaner.

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,55 @@
...
+            - name: Attach VSIX package to release
+              uses: actions/github-script@v6
+              with:
+                  script: |
+                      const fs = require('fs');
+                      const packageJson = require('./package.json');
+                      const packageName = packageJson.name;
+                      const packageVersion = packageJson.version;
+                      const vsixFilename = `${packageName}-${packageVersion}.vsix`;
+                      const uploadUrl = await github.rest.repos.uploadReleaseAsset({
```
sfdctaka, line 49: nit: Do we need `uploadUrl` if we're not returning its property, something like we co in `Create Release` with `return response.data.id`?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,55 @@
...
+                      const packageVersion = require('./package.json').version;
+                      const response = await github.rest.repos.createRelease({
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
+                        tag_name: `v${packageVersion}`,
+                        name: `Release ${packageVersion}`,
+                        body: `Generating release for version ${packageVersion}`
+                      });
+                      return response.data.id;
+            - name: Attach VSIX package to release
```
sfdctaka, line 40: nit nit: Missing newline since other steps are delineated by single line space in between.

sfdctaka (approved):
LGTM!

.github/workflows/create_release.yml
khawkins: Haha I was just counting on the code formatting to fix that kind of stuff for me! I guess not. 😆 Good catch!

.github/workflows/create_release.yml
khawkins: Ah, good catch. No. I had started this script anticipating more steps to come after, then didn't do that. I'll take it out.

khawkins:
Gonna merge this and see if it works. It worked in a sandbox environment I used for coding it up, so 🤞. Happy to make iterations as we go.

RELEASE.md
```diff
@@ -1,25 +1,25 @@
...
-3. Create package
+2.  Create the VSIX package
 
-```
-vsce package
-```
+        npx vsce package
 
-4. Manually install the generated .vsix file and smoke test.
+    **NOTE:** The packaging workflow prunes the dev dependencies out of the `node_modules/` tree, so you'll need to re-run `npm install` to reinstall dev dependencies and continue development work.
```
dbreese, line 17: Do you want to remove this since other PR removes the pruning now?

khawkins: Oh thanks. 😊 Yeah, I'll remove it.

khawkins: Oops I just realized I already merged this. I'll get it in the open one!

PR #33, opened 2023-06-30
Pruning dev is breaking vsce packaging

khawkins:
The size of the bundle doesn't change, and doing the dev pruning is otherwise breaking `vsce` when installed as a dev dependency.

Let's save the file culling for when we optimize the package.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,8 +14,6 @@ The following instructions will create the extension package (`salesforcedx-vsco
 
         npx vsce package
 
-    **NOTE:** The packaging workflow prunes the dev dependencies out of the `node_modules/` tree, so you'll need to re-run `npm install` to reinstall dev dependencies and continue development work.
-
 3.  Install the generated .vsix file and test as you will:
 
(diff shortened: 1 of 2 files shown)


sfdctaka (approved):
LGTM!

dbreese:
Interesting, I was seeing it change in sizes, so perhaps it was something different causing the file to change sizes.

dbreese (approved):

dbreese:
Probably want to update the RELEASE.md to remove the section you added about the pruning as well.

khawkins:
> Probably want to update the RELEASE.md to remove the section you added about the pruning as well.

Yep I'll update that over here. Good catch!

khawkins:
> Yip, you are correct, sir! Not sure what else I was seeing, but maybe it was a combination of trying the `-external:*` parms on esbuild and a lot of other things. Sorry about that!

Oh no worries, there are a _lot_ of moving parts in the bundling and packaging. It's taken some time for me to even get my head 75% of the way around it. The remaining 25% is still 🫤.

PR #36, opened 2023-07-06
Install package, to use vsce dependency for publish

khawkins:
I noticed in the publish flow that `npx vsce publish` downloaded and installed `vsce` (and complained about not using its full package name), because we weren't installing the package dependencies of the project.


khawkins:
(Foregoing a version bump, since the change is entirely outside the package. This can get pulled into the next release.)

dbreese (approved):

sfdctaka (approved):
LGTM!

PR #43, opened 2023-09-07
Fixing Node engine, Volta versions

khawkins:
Volta does not support version ranges on pinning, so using Volta with the previous `>=18` version was actually breaking for Volta installations.

Moved our Node engine minimum version to the `engines` configuration, and pinned Node 18 for Volta at the latest version.


sfdctaka (approved):
LGTM!

PR #44, opened 2023-10-18
Move code around

maliroteh-sf:
Move code for commands to subfolders. Subfolder names are category names that the commands belong to, which will help us keep code organized as we add more commands in the future:
```
src/commands/wizard/...
src/commands/utam/...
src/commands/lwc/...
etc.
```


src/commands/wizard/onboardingWizard.ts
```diff
@@ -0,0 +1,94 @@
+/*
```
maliroteh-sf, line 1: Moved most of the code from `extension.ts` to here

khawkins (approved):
Looks good!

PR #46, opened 2023-10-18
Webviews: Two-way messaging

khawkins:
In order to support the logic of the forthcoming landing page selection and LWC QA generation, we needed a two-way channel between the webviews and their backing TypeScript controller pages. This does that. I took out the previous one-way button event messaging, and replaced it with a more general two-way protocol.

This is feature complete, but needs tests. But I wanted to get the concept in front of you all for any feedback you wish to offer.


resources/instructions/landingpage.html
```diff
@@ -1,19 +1,29 @@
-<!DOCTYPE html>
```
khawkins, line 1: I wouldn't give this page too much scrutiny, as it's going away imminently. I just updated for completeness sake.

src/commands/wizard/authorizeCommand.ts
```diff
@@ -24,7 +24,7 @@ export class AuthorizeCommand {
             if (!result || result.title === l10n.t('No')) {
                 return Promise.resolve(false);
             } else {
-                await commands.executeCommand('sfdx.force.auth.web.login');
+                await commands.executeCommand('sfdx.org.login.web');
```
khawkins, line 27: This is actually broken in the current code. 😔 Apparently the command names moved on and we didn't know it.

src/webviews.ts
```diff
@@ -40,12 +47,24 @@ export class InstructionsWebviewProvider {
...
+            );
+            if (responsiveHandlers.length > 0) {
+                const handler = responsiveHandlers[0];
+                let callback: WebviewMessageCallback | undefined;
+                if (data.callbackId) {
+                    const returnedCallbackId = data.callbackId;
+                    delete data.callbackId;
+                    callback = (responseData?: object) => {
+                        const fullResponseMessage = {
+                            callbackId: returnedCallbackId,
```
khawkins, line 61: The magic glue!

dbreese:
Approach looks good to me!

maliroteh-sf (approved):

sfdctaka (approved):
LGTM!

khawkins:
Got good coverage on the webview helper class I pulled out (`WebviewProcessor`), and left testing of `InstructionsWebviewProvider` alone, because what's left there doesn't really do much but hand off to `WebviewProcessor`.

Still need to look at how I can test the JS side of the messaging, but that's for another day.

src/webviews/processor.ts
```diff
@@ -0,0 +1,127 @@
...
+    public onWebviewReceivedMessage(
+        data: any,
+        panel: WebviewPanel,
+        messageHandlers: WebviewMessageHandler[]
+    ) {
+        const responsiveHandlers = messageHandlers.filter(
+            (messageHandler) => data.type === messageHandler.type
+        );
+        if (responsiveHandlers.length > 0) {
+            const handler = responsiveHandlers[0];
```
sfdctaka, line 69: nit: Refactor into a new file(processor.ts) hides the assertion that why index 0 handler is a valid handler. Since the validation of handler map takes place at all the caller(`showInstructionWebview`) maybe `onWebviewReceivedMessage` should be passed a single message handler instead of an array there?

khawkins: That's a good catch! In my next PR I'm actually moving the handler validation into the processor as well, for the reason you stated and the reason that it's really more congruent there (another aspect you pointed out).

While I like having the array just passed in here for its own processing, I do agree that the selection of just the first item feels a little oblique. I could either add a comment here, or what I'm leaning toward is just switching to `messageHandlers.find()` instead, which only returns the first item found (which is always going to be 0 or 1 items), and probably reads better for this use case. Let me know what you think.

PR #47, opened 2023-10-26
Upgrade dependencies to latest

maliroteh-sf:


package.json
```diff
@@ -53,36 +53,36 @@
     "test-coverage": "node ./out/test/runTest.js --coverage",
     "prettier:write": "prettier --write \"src/**/*.{ts, js}\"",
     "prettier:verify": "prettier --list-different \"src/**/*.{ts, js}\"",
-    "bundle:extension": "esbuild ./src/extension.ts  --bundle --outdir=out --format=cjs --target=es2020 --platform=node --external:vscode --external:@oclif/core --minify --sourcemap",
+    "bundle:extension": "esbuild ./src/extension.ts --bundle --outdir=out --format=cjs --target=es2020 --platform=node --external:vscode --external:@oclif/core --external:@salesforce/lwc-dev-mobile-core --minify --sourcemap",
```
maliroteh-sf, line 56: Starting with v3.2.0 of `lwc-dev-mobile-core`, it has upgraded its dependency on `@salesforce/core` from `3.x` to `5.x` This seems to have changed the behavior of `@salesforce/core` around loading message files for loggers.

Since we're creating a bundle, if we don't exclude `lwc-dev-mobile-core` then it won't be able to load its message files. This is b/c `lwc-dev-mobile-core` has code like `Messages.importMessagesDirectory(__dirname);` but if it is being bundled, then wrong value will be used for `__dirname` (value would be the path to the extension root dir rather than the path to the dependency root dir in `node_modules`)

khawkins (approved):
Awesome! Thanks for catching this.

PR #48, opened 2023-10-27
Choose landing page from templates

khawkins:
Allows the user to choose their landing page from one of the templates in the Offline Starter Kit project.

### DIFF
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -3,7 +3,7 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-        <title>Offline Starter Kit: Create SObject LWC Quick Actions</title>
+        <title>Offline Starter Kit: Create sObject LWC Quick Actions</title>
     </head>
     <style>
         h1 {
@@ -27,7 +27,7 @@
         }
     </style>
     <body>
-        <h1>Create SObject LWC Quick Actions</h1>
+        <h1>Create sObject LWC Quick Actions</h1>
 
         <p>
             The following sObjects are present in your configured landing page:
@@ -36,7 +36,7 @@
         <table>
             <thead>
                 <tr>
-                    <th>SObject</th>
+                    <th>sObject</th>
                     <th>LWC Quick Actions</th>
                 </tr>
             </thead>
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -1,0 +1,71 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Offline Starter Kit: Create SObject LWC Quick Actions</title>
+    </head>
+    <style>
+        h1 {
+            font-size: 16px;
+        }
+        h2 {
+            font-size: 14px;
+        }
+        table {
+            border-collapse: collapse;
+        }
+        table,
+        th,
+        td {
+            border-width: 1px;
+            border-style: solid;
+        }
+        th,
+        td {
+            text-align: center;
+        }
+    </style>
+    <body>
+        <h1>Create SObject LWC Quick Actions</h1>
+
+        <p>
+            The following sObjects are present in your configured landing page:
+        </p>
+
+        <table>
+            <thead>
+                <tr>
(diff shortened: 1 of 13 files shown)


resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -0,0 +1,81 @@
+<!doctype html>
```
khawkins, line 1: This one's just a mock so far. It won't be called in the current workflow.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -0,0 +1,30 @@
+import { Uri, l10n } from 'vscode';
+import { InstructionsWebviewProvider } from '../../webviews/instructions';
+
+export class LwcGenerationCommand {
```
khawkins, line 4: Another one that's just stubbed. Not accessible to the user yet.

resources/instructions/landingPageTemplateChoice.html
```diff
@@ -0,0 +1,262 @@
...
+                    chooseTemplateButton.disabled = true;
+                } else {
+                    for (const landingPageType in response.landingPageCollection) {
+                        const landingPageStatus =
+                            response.landingPageCollection[landingPageType];
+                        if (!landingPageStatus.exists) {
+                            const landingPageLabelId = `${landingPageType}LandingPageTemplateLabel`;
+                            const landingPageLabelElement =
+                                document.getElementById(landingPageLabelId);
+                            if (landingPageStatus.warning) {
```
dbreese, line 224: I'm not understanding what this is for as I dont see any warning text defined anywhere?

resources/instructions/landingPageTemplateChoice.html
```diff
@@ -0,0 +1,262 @@
...
+                        <p class="landing-page-option-title">Keep Existing</p>
+                        <p class="landing-page-option-subtitle">
+                            Keep the landing page that's already configured in
+                            your project.
+                        </p>
+                    </div>
+                </div>
+            </label>
+
+            <label id="defaultLandingPageTemplateLabel">
```
dbreese, line 64: So adding new landing pages in the future will require copy-pasting the html snippets here in this file?

dbreese (approved):
Some questions, but LGTM

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -0,0 +1,81 @@
...
+        th,
+        td {
+            text-align: center;
+        }
+    </style>
+    <body>
+        <h1>Create SObject LWC Quick Actions</h1>
+
+        <p>
+            The following sObjects are present in your configured landing page:
```
sfdctaka, line 33: nit: Probably standardize on SObjects? I'm not sure which capitalization is actually preferred but SObject seems to be used throughout.

sfdctaka (approved):
LGTM!

resources/instructions/createSObjectLwcQuickActions.html
khawkins: I had done it as a distinction between title case (headings) and lower case (in sentences). But I think for title case, I erroneously looked at [the `SObject` class](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_methods_system_sobject.htm) in the Apex Reference Guide, which is likely taking casing queues based more on coding standards than nomenclature.

The [Use sObjects](https://trailhead.salesforce.com/content/learn/modules/apex_database/apex_database_sobjects) trail in Trailhead seems to treat it as "sObject" in all cases. So that's probably the one worth standardizing on. Good callout!

resources/instructions/landingPageTemplateChoice.html
khawkins: At the moment. But it's a good point: all of these entries could be dynamically generated based on returned data—which they already get, to drive their visibility.

I had thought about that approach, and at the end of the day wasn't sure how much the entries in this page would fluctuate, to make it worthwhile to drive the content dynamically. We're well positioned to do so though, if we decide to go that route.

resources/instructions/landingPageTemplateChoice.html
khawkins: The warning data comes back as one of the status data items from the `TemplateChooserCommand.getLandingPageStatus()` [method](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/48/files#diff-7d148ba5f620ba14e46a900655961c4d76f535280279cdb20e203adbbe52bf45R228-R230). That status check will create a warning if one or both of the landing page files for the given template are not available in the project.

PR #50, opened 2023-11-07
Checks for a list of sobjects and the corresponding create/edit/view qa directories.

dbreese:
May require some tweaking during integration, but this should be a good stepping stone.

@sfdctaka @khawkins @maliroteh-sf

### DIFF
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -1,81 +1,110 @@
 <!doctype html>
 <html lang="en">
-    <head>
-        <meta charset="UTF-8" />
-        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-        <title>Offline Starter Kit: Create sObject LWC Quick Actions</title>
-    </head>
-    <style>
-        h1 {
-            font-size: 16px;
-        }
-        h2 {
-            font-size: 14px;
-        }
-        table {
-            border-collapse: collapse;
-        }
-        table,
-        th,
-        td {
-            border-width: 1px;
-            border-style: solid;
-        }
-        th,
-        td {
-            text-align: center;
-        }
-    </style>
-    <body>
-        <h1>Create sObject LWC Quick Actions</h1>
 
-        <p>
-            The following sObjects are present in your configured landing page:
-        </p>
+<head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Offline Starter Kit: Create sObject LWC Quick Actions</title>
+</head>
+<style>
+    h1 {
+        font-size: 16px;
+    }
 
-        <table>
-            <thead>
-                <tr>
-                    <th>sObject</th>
-                    <th>LWC Quick Actions</th>
-                </tr>
-            </thead>
-            <tbody>
-                <tr>
-                    <td>Contact</td>
-                    <td>✅</td>
-                </tr>
-                <tr>
-                    <td>Account</td>
-                    <td>✅</td>
-                </tr>
-                <tr>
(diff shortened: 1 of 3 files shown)


dbreese:
Hmmm, not sure why "templateChooserCommand" test is failing.

dbreese:
Appears to possibly be a flapper. I just re-ran the failed test and it passed with no issues, so we can keep an eye on ie.

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -1,81 +1,110 @@
 <!doctype html>
 <html lang="en">
-    <head>
```
dbreese, line 3: Sorry, most of the changes is just the addition of the scripts in the html. I made mistake of reformatting doc without realizing.

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -1,81 +1,110 @@
...
-                    'generateLwcQuickActionsButton'
-                );
-            });
-        </script>
-        <script src="--- MESSAGING_SCRIPT_SRC ---"></script>
-    </body>
-</html>
+        });
+
+        // Wait until all scripts are loaded, before engaging with e.g.
```
dbreese, line 76: This is the meat of the change here -- makes a call to new function and dynamically generates table rows.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -22,9 +35,81 @@ export class LwcGenerationCommand {
                             panel.dispose();
                             return resolve();
                         }
+                    },
+                    {
+                        type: 'getQuickActionStatus',
+                        action: async (_panel, data, callback) => {
+                            // TODO: Hook this up to function from W-14439417 that parses landing_page.json.
+                            const sobjects = [
```
dbreese, line 43: Just a hard-coded list for now.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -22,9 +35,81 @@ export class LwcGenerationCommand {
                             panel.dispose();
                             return resolve();
                         }
+                    },
+                    {
+                        type: 'getQuickActionStatus',
+                        action: async (_panel, data, callback) => {
```
khawkins, line 41: Nit: `data` will probably go to `_data` too, unless you anticipate passing any in.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -22,9 +35,81 @@ export class LwcGenerationCommand {
...
             );
         });
     }
+
+    static async checkForExistingQuickActions(
+        sobjects: string[]
+    ): Promise<SObjectQuickActionStatus> {
+        return new Promise<SObjectQuickActionStatus>(async (resolve) => {
+            const results: SObjectQuickActionStatus = { sobjects: {} };
+            results.sobjects = {};
```
khawkins, line 68: You don't need this, right? looks like it's initialized as such in your declaration.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -22,9 +35,81 @@ export class LwcGenerationCommand {
...
+
+            return resolve(results);
+        });
+    }
+
+    private static checkForExistingQuickAction(
+        sobject: string,
+        qaName: string
+    ): boolean {
+        const expectedDirName = `${sobject}.${qaName}.quickAction-meta.xml`;
```
khawkins, line 103: Nit: this variable name could probably be slightly more appropriate, e.g. `expectedMetadataFilename` or somesuch.

khawkins (approved):
Approach looks good to me overall! Certainly in line with what we did for landing page template selection.

PR #51, opened 2023-11-08
Parse sObjects in the landing page and return them in an array

sfdctaka:


src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -8,6 +18,70 @@ export class LwcGenerationCommand {
...
+        filePath: string,
+        callback: (error: Error | null, data: any) => void
+    ): void {
+        try {
+            const data = fs.readFileSync(filePath, { encoding: 'utf-8' });
+            const jsonObject = JSON.parse(data);
+            callback(null, jsonObject);
+        } catch (error) {
+            callback(error as Error, null);
+        }
```
sfdctaka, line 31: Is Promise preferred over callback?

khawkins:
You're going to need to rebase against `main`.

src/utils/uemParser.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import { l10n } from 'vscode';
+import { Field } from './orgUtils';
+
+export class UEMParser {
+    public static findFieldValues(json: Object, targetField: String): string[] {
+        const results: string[] = [];
+
+        function search(json: any) {
```
khawkins, line 15: Why do you have a function within a function here? I would either inline all of this functionality, or pull out a helper method from the wrapping method, so that it can be more easily tested in isolation.

src/commands/wizard/lwcGenerationCommand.ts
dbreese: I think Promise should be used any time there could be a blocking operation as in the above code where you are reading the file synchronously. If the consumer of readFileAsJsonObject() doesn't know this, it can have negative side affects. Whereas using a Promise automatically indicates it can be async.

sfdctaka:
@khawkins On resolving the conflict now. I was talking with @dbreese but after resolving we will need a refinement in how we render the table as we'll now have both what sObjects that we are generating for and what sObjects' QAs already exist. So this may not come clear after my merge conflict, just a heads up.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -8,6 +18,70 @@ export class LwcGenerationCommand {
         this.extensionUri = extensionUri;
     }
 
+    static readFileAsJsonObject(
```
maliroteh-sf, line 21: We are already pulling in `lwc-dev-mobile-core` into this repo and so you can just use `CommonUtils.loadJsonFromFile` to get the job done, which also means that you don't need to add test coverage for this.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -8,6 +18,70 @@ export class LwcGenerationCommand {
...
+                    landingPagePath,
+                    (error: Error | null, data: any) => {
+                        if (error) {
+                            console.warn(`Error reading ${landingPageJson}`);
+                            lwcGenerationCommandStatus.error = (
+                                error as Error
+                            ).message;
+                        } else {
+                            lwcGenerationCommandStatus.sobjects =
+                                UEMParser.findFieldValues(
```
dbreese, line 71: I am wondering if we want to find _all_ "objectApiName" keys in the entire json, or just for the mcf/list, mcf/timedList, and mcf/genericLists? 

Because the purpose of generating the LWCs for view/edit/create is that the lists are the jumping point for sobjects. Looks like the only place objectApiName appears in the templates is under either `mcf/list` or `mcf/timedList` so this should still work.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -8,6 +18,70 @@ export class LwcGenerationCommand {
...
+            callback(error as Error, null);
+        }
+    }
+
+    static async getLwcGenerationPageStatus(): Promise<LwcGenerationCommandStatus> {
+        return new Promise<LwcGenerationCommandStatus>(async (resolve) => {
+            let landingPageExists = true;
+
+            const staticResourcesPath =
+                await TemplateChooserCommand.getStaticResourcesDir();
```
khawkins, line 39: @dbreese and I (coincidentally) discussed this earlier: these commands should move out into independent helper functionality. I'd prefer we do that, over importing `TemplateChooserCommand` to get to them.

src/commands/wizard/lwcGenerationCommand.ts
khawkins: +1. I'd prefer an approach where, rather than blindly searching the landing page, we go the other way and get values based on known paths to those keys.

src/commands/wizard/onboardingWizard.ts
```diff
@@ -33,6 +33,10 @@ async function runPostProjectConfigurationSteps(
         await AuthorizeCommand.authorizeToOrg();
         await DeployToOrgCommand.deployToOrg();
 
+        await new LwcGenerationCommand(
```
khawkins, line 36: Are we ready to enable this functionality, after this PR goes in? If not, we should not put this in the PR.

khawkins:
> @khawkins On resolving the conflict now. I was talking with @dbreese but after resolving we will need a refinement in how we render the table as we'll now have both what sObjects that we are generating for and what sObjects' QAs already exist. So this may not come clear after my merge conflict, just a heads up.

Yeah we need to resolve, for sure. Looking at both implementations, I don't think the results of your functionality (finding the sObjects represented in the landing page) should be going back to the UI. That's functionality that the LWC generation code needs to invoke, to understand what it needs to generate.  To me, that's what should be (and currently is) going back to the UI. I think the sObject interrogation in the landing page should just be represented as functionality internal to the process.

Cc: @dbreese

src/commands/wizard/onboardingWizard.ts
sfdctaka: After this PR goes in the remaining thing will be to get the compact layout. We can disable the functionality until that is in.

sfdctaka:
@dbreese @khawkins I am now working on exactly that, making interrogation as a part or internal to finding which sobject's QA to generate(view/edit/create).  The way it is set up now `window.load` will need to make to separate calls for finding what sobjects there are on the landing page and which QA's are there for those sobjects. Mashing them into one is the way to go.

src/commands/wizard/onboardingWizard.ts
sfdctaka: Removed it. Will continue to put the command on dev's machine locally until we're ready to enable the functionality. cc: @dbreese

package.json
```diff
@@ -58,6 +58,7 @@
   },
   "devDependencies": {
     "@istanbuljs/nyc-config-typescript": "^1.0.2",
+    "@types/cli-progress": "^3.11.5",
```
sfdctaka, line 61: To use `CommonUtils.loadJsonFromFile` I needed to include this or otherwise compiler complains @maliroteh-sf

dbreese (approved):
LGTM!

package.json
khawkins: That's super weird. I don't know why the specific use of that method is triggering the issue.

The dependency tree for that package is entirely housed in an oclif _dev_ dependency in `lwc-dev-mobile-core`. It's a little disconcerting that we're dragging all of OClif along for the essentially generic Node.js utility functions that we leverage from `lwc-dev-mobile-core`. I'm taking a note to revisit this phenomenon in that package.

For the time being, the solution you chose is probably the easiest—that, or not using the method in favor of writing a local one in this project. At this time I don't know how we'd actually clear that error in the place it rightfully should be cleared.

src/commands/wizard/lwcGenerationCommand.ts
khawkins: I see this is resolved, but...can we do this? 😀 Is that TBD?

src/utils/uemParser.ts
```diff
@@ -0,0 +1,69 @@
...
+        return results;
+    }
+
+    static findObjectsWithValues(
+        json: Object,
+        valuesToMatch: string[]
+    ): Array<Object> {
+        const results: Array<Object> = [];
+
+        if (typeof json === 'object') {
```
khawkins, line 38: This is a redundant check, to the required type in the method signature. I'd remove the conditional.

src/utils/uemParser.ts
```diff
@@ -0,0 +1,69 @@
...
+            // Only include unique values in the array.
+            if (!results.includes(objectApiName)) {
+                results.push(objectApiName);
+            }
+        });
+
+        return results;
+    }
+
+    static findObjectsWithValues(
```
khawkins, line 32: Sorry, I'm still not convinced on "searching" the JSON. The JSON in question in this case is a well-known data structure. We should be querying it, not searching it. For all of its unwieldiness, it's still deterministic that `objectApiName` keys can be found at `view.regions.components.components[_x_].regions.components.components[_y_].properties.objectApiName` (or something along those lines). I haven't looked across all of the templates to see whether or not there may be a similar path to other ones—but if there is, it will be just as deterministic.

I feel like we should be deliberate about access these data properties. To me, searching for them is just opening a door to false positives, with no real upside. We should query the data structure we know.

dbreese: I wasn't sure if it will always be in that location. I suggested to @sfdctaka that as long as we find an `mcf/list` or `mcf/timedList`, then from there `.properties.objectApiName` should be populated. I thought that with the UEM file, it can be arbitrarily deep and nested in other components such as an `mcf/card` or something else perhaps.

package.json
maliroteh-sf: `CommonUtils` has 3 methods `startCliAction` / `updateCliAction` / `stopCliAction` that import `ux` from `@oclif/core`

I guess when trying to use `CommonUtils.loadJsonFromFile` the entire `CommonUtils` is pulled in and so you end up needing to add this dependency. In `lwc-dev-mobile-core` we have a dependency on `@oclif/core` which in turn pulls in `@types/cli-progress` so I guess what you're doing here is ok.

src/utils/uemParser.ts
khawkins: I support continuing to have some measure of a UEM "fronting" class for these data structures, because they are so unwieldy, and accessing properties therein in a more friendly encapsulated manner could be useful down the road.

khawkins: > I thought that with the UEM file, it can be arbitrarily deep and nested in other components such as an `mcf/card` or something else perhaps.

It'd be good to understand if that's the case, and then if so what the constraints are on that definition, and how we'd respond to them.

My point is more that we should strive to leverage our understanding of this data structure that we're supporting. To do a "find" is kind of punting on that understanding, and to me exposes a gap that we need to close.

package.json
khawkins: I wonder why it's pulling in the entirety of `CommonUtils`, when other methods aren't. I wonder if it's because `loadJsonFromFile` returns `any`, maybe?

src/utils/uemParser.ts
sfdctaka: cc: @Taost

sfdctaka: I wrote a xpath-ish query function to replace the recursive deep search. @khawkins please check it out.

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -35,76 +35,76 @@
...
-    <button id="generateLwcQuickActionsButton">
-        Generate LWC Quick Actions
-    </button>
-    <script>
-        const generateQuickActionsButtonElement = document.getElementById(
-            'generateLwcQuickActionsButton'
-        );
-        generateQuickActionsButtonElement.addEventListener('click', () => {
-            webviewMessaging.sendMessageRequest(
+        <p id="sobjectDiv">
```
khawkins, line 38: Looking through this, I'm realizing that all of the HTML files under `resources/instructions` have nominally been put through `prettier` formatting already. Let's codify that as part of this PR:

- Run `prettier` on this file, to fix up spacing, strings, etc.
- Amend `prettier:verify` in `package.json` to add the `"resources/instructions/**/*.html"` path.
- Amend `prettier:write` to do the same.

We may as well try to keep these formatted too, since they mostly already are.

dbreese: I think this is a great idea. HOWEVER, can we please not do it in this PR because I have stacked changes on this PR and will have to go through even more merge conflicts if we do. I can add this in my next PR?

khawkins: Oh yeah sure. No need to blow up the train as part of the process!

sfdctaka: Noted. I will make another PR for `prettier` after this gets merged.

src/utils/uemParser.ts
```diff
@@ -0,0 +1,68 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export class UEMParser {
+    public static findSObjects(json: Object): Array<string> {
+        const sObjects = UEMParser.findObjectsWithValues(json, [
+            'mcf/list',
+            'mcf/timedList',
+            'mcf/genericLists'
```
Taost, line 13: There is no 'mcf/genericLists' definition in MCF. The 'mcf/list' is basically for generic list.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -9,18 +14,52 @@ export type QuickActionStatus = {
...
 export class LwcGenerationCommand {
     extensionUri: Uri;
 
     constructor(extensionUri: Uri) {
         this.extensionUri = extensionUri;
     }
 
+    static async getSObjectsFromLandingPage(): Promise<GetSObjectsStatus> {
+        return new Promise<GetSObjectsStatus>(async (resolve) => {
+            const staticResourcesPath = await UIUtils.getStaticResourcesDir();
```
Taost, line 37: It looks good if this tool is specifically for landing page coming from static resources. But, landing page uem can also come from other sources, e.g., mobile builder, the default (packaged inside app).

sfdctaka:
@khawkins Made the search to simple look for all `objectApiName`. I will do `prettier` update in another PR.

src/utils/uemParser.ts
```diff
@@ -0,0 +1,47 @@
...
+                } else {
+                    results.push(
+                        ...UEMParser.findObjectsWithKey(
+                            nestedJsonBlock[key as keyof Object],
+                            keyToMatch
+                        )
+                    );
+                }
+            }
+        } else if (Array.isArray(nestedJsonBlock)) {
```
khawkins, line 37: I don't see a use case where this is going to happen. A JS/TS array is `typeof object`.

src/utils/uemParser.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    static findObjectsWithKey(
+        nestedJsonBlock: any,
+        keyToMatch: string
+    ): Array<string> {
+        const results: Array<any> = [];
+
+        if (typeof nestedJsonBlock === 'object') {
+            const keys = Object.keys(nestedJsonBlock);
```
khawkins, line 22: You don't need this. It's not used.

src/utils/uemParser.ts
khawkins: Plus, you don't need to special-case arrays from the logic you have here. `for...in` will return the properties of a JSON object, and the indices of an array.  For example:

```js
let myArray = ["One", "Two", "Three"];
for (const arrayKey in myArray) {
  console.log(`${arrayKey}: ${myArray[arrayKey]}`);
}

// Output:
// 0: One
// 1: Two
// 2: Three
```

src/utils/uiUtils.ts
```diff
@@ -5,12 +5,25 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { window, QuickPickItem, QuickPickItemKind, QuickPick } from 'vscode';
+import { window, workspace, QuickPickItem } from 'vscode';
+import { access } from 'fs/promises';
+import {
```
khawkins, line 10: These error objects need to be moved into this module as well.

khawkins: Actually, can we move them all into their own module, somewhere separate from `uiUtils`? They're not really UI utilities. Something like `WorkspaceUtils` would probably be more appropriate.

src/utils/uiUtils.ts
```diff
@@ -64,4 +77,39 @@ export class UIUtils {
...
+            const staticResourcesPath = path.join(
+                projectPath,
+                this.STATIC_RESOURCES_PATH
+            );
+            try {
+                await access(staticResourcesPath);
+            } catch (err) {
+                const accessErrorObj = err as Error;
+                const noAccessError = new NoStaticResourcesDirError(
+                    `Could not read landing page directory at '${staticResourcesPath}': ${accessErrorObj.message}`
```
khawkins, line 108: Nit: we should probably change this error message to something more generic, now that it's split out into a utility function. E.g.:

> Could not read static resources directory at '${staticResourcesPath}'

khawkins (commented):
It looks good to me overall! A few minor corrections, but otherwise it's lookin' good.

dbreese (approved):

src/utils/uiUtils.ts
```diff
@@ -64,4 +74,39 @@ export class UIUtils {
             quickPick.show();
         });
     }
+
+    static getWorkspaceDir(): string {
```
khawkins, line 78: Sorry, to be clear: I think this method, `STATIC_RESOURCES_PATH` above, and `getStaticResourcesDir()` below should also end up in `workspaceUtils`, as none of those are particularly related to "UI". Can we move these over as well?

sfdctaka:
I think the last commit could be the final dagger, @khawkins .

src/test/suite/utils/uiUtils.test.ts
```diff
@@ -94,4 +102,62 @@ suite('UIUtils Test Suite', () => {
         assert.equal(exceptionCount, 1);
         assert.equal(quickPickSpy.dispose.called, true); // ensure it was disposed of after item selected
     });
+
+    test('Static resources dir: workspace does not exist', async () => {
```
khawkins, line 106: You can remove these tests from this file, right? Looks like you (rightly) moved them into `workspaceUtils.test.ts` below.

khawkins:
> I think the last commit could be the final dagger, @khawkins .

Pretty close! 😆 I think all that's left is to just remove those now-redundant tests in `uiUtils.test.ts`. Otherwise 🧑‍🍳😚 looking good to me!

khawkins (approved):
🎉

src/test/suite/utils/uiUtils.test.ts
sfdctaka: So strange because I removed them... Checked now and saw that `git push` hadn't gone through because I didn't add `-f` to it when the removal of those tests were done in ammend. 😞 . So much for haste makes waste.

PR #52, opened 2023-11-08
Now able to generate LWCs using template files.

dbreese:
@sfdctaka

### DIFF
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -84,18 +84,12 @@
 
         generateQuickActionsButtonElement.addEventListener('click', () => {
             webviewMessaging.sendMessageRequest(
-                'generateLwcQuickActionsButton',
+                'generateLwcQuickActions',
                 {},
                 handleQuickActionStatusResponse
             );
         });
         
-        generateQuickActionsButtonElement.addEventListener('click', () => {
-            webviewMessaging.sendMessageRequest(
-                'generateLwcQuickActionsButton'
-            );
-        });
-
         continueButtonElement.addEventListener('click', () => {
             webviewMessaging.sendMessageRequest(
                 'continueButton'
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -56,18 +56,30 @@
             </tbody>
         </table>
 
-        <label
-            style="display: block; margin-top: 10px; margin-bottom: 8px"
-            for="generateLwcQuickActionsButton"
-        >
-            Click 'Generate LWC Quick Actions' to create the missing LWC Quick
-            Actions:
-        </label>
-        <button id="generateLwcQuickActionsButton">
-            Generate LWC Quick Actions
-        </button>
-        <script>
-            const generateQuickActionsButtonElement = document.getElementById(
+    <label style="display: block; margin-top: 10px; margin-bottom: 8px" for="generateLwcQuickActionsButton">
+        Click 'Generate LWC Quick Actions' to create the missing LWC Quick
+        Actions:
+    </label>
+
+    <button id="generateLwcQuickActionsButton">
+        Generate LWC Quick Actions
+    </button>
+
+    <button id="skipStepButton">
+        Skip This Step
+    </button>
+
+    <script>
+        const generateQuickActionsButtonElement = document.getElementById(
+            'generateLwcQuickActionsButton'
+        );
+
+        const skipStepButtonElement = document.getElementById(
+            'skipStepButton'
+        );
+
+        generateQuickActionsButtonElement.addEventListener('click', () => {
+            webviewMessaging.sendMessageRequest(
                 'generateLwcQuickActionsButton'
             );
             generateQuickActionsButtonElement.addEventListener('click', () => {
@@ -76,16 +88,21 @@
                 );
             });
 
-            // Wait until all scripts are loaded, before engaging with e.g.
-            // messaging functionality.
-            window.addEventListener('load', () => {
-                // What's the status of the various sobject quick actions in the landing_page.json?
-                webviewMessaging.sendMessageRequest(
-                    'getQuickActionStatus',
-                    {},
-                    handleQuickActionStatusResponse
-                );
-            });
+        skipStepButtonElement.addEventListener('click', () => {
+            webviewMessaging.sendMessageRequest(
+                'skipStepButton'
+            );
+        });
+
+        // Wait until all scripts are loaded, before engaging with e.g.
+        // messaging functionality.
+        window.addEventListener('load', () => {
+            // What's the status of the various sobject quick actions in the landing_page.json?
+            webviewMessaging.sendMessageRequest(
+                'getQuickActionStatus',
+                {},
+                handleQuickActionStatusResponse
+            );
 
             function handleQuickActionStatusResponse(response) {
                 const table = document.getElementById('quickActionStatusTable');
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -84,11 +84,13 @@
                 {},
                 handleQuickActionStatusResponse
             );
-            generateQuickActionsButtonElement.addEventListener('click', () => {
-                webviewMessaging.sendMessageRequest(
-                    'generateLwcQuickActionsButton'
-                );
-            });
+        });
+        
+        generateQuickActionsButtonElement.addEventListener('click', () => {
+            webviewMessaging.sendMessageRequest(
+                'generateLwcQuickActionsButton'
+            );
+        });
 
         continueButtonElement.addEventListener('click', () => {
             webviewMessaging.sendMessageRequest(
@@ -105,11 +107,8 @@
                 {},
                 handleQuickActionStatusResponse
             );
-
-            function handleQuickActionStatusResponse(response) {
-                const table = document.getElementById('quickActionStatusTable');
-                for (const sobject in response.sobjects) {
-                    const quickActions = response.sobjects[sobject];
+            
+        });
 
         function handleQuickActionStatusResponse(response) {
             const table = document.getElementById('quickActionStatusTable');
@@ -135,6 +134,7 @@
                 edit.innerHTML = quickActions.edit == true ? "✅" : "❌";
                 create.innerHTML = quickActions.create == true ? "✅" : "❌";
             }
+        }
         </script>
         <script src="--- MESSAGING_SCRIPT_SRC ---"></script>
     </body>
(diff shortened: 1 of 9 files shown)


dbreese:
Ah, Windows is failing due to path separator issues. I will fix in update in a few minutes.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -139,6 +152,52 @@ export class LwcGenerationCommand {
...
+            for (const sobject in quickActionStatus.sobjects) {
+                const quickActions = quickActionStatus.sobjects[sobject];
+
+                if (
+                    !quickActions.create ||
+                    !quickActions.edit ||
+                    !quickActions.view
+                ) {
+                    // at least 1 needs to be created
+                    const compactLayoutFields =
```
dbreese, line 169: This is expensive, so only doing it if we know we need to generate something.

src/test/suite/utils/orgUtils.test.ts
```diff
@@ -165,17 +165,20 @@ suite('Org Utils Test Suite', () => {
                 {
                     editableForNew: true,
                     editableForUpdate: true,
-                    label: 'Name'
+                    label: 'Name',
```
dbreese, line 168: We could technically remove editableForNew/Update and this label value, because we thought we were going to use it and so far dont think we need to.

We still need to do some testing. It appears that the base LWC components used in our view/edit/create templates checks these values for us.

However, if we include "Owner", for example which has editableForNew=false and editableForUpdate=false, on the edit and create LWCs, it will still be shown, but will be disabled. Functionality still works fine. I think for the CREATE case, this may be odd because it will be a blank field shown, but still disabled. For the EDIT case, it still makes sense to show the value.

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -56,54 +56,87 @@ <h1>Create sObject LWC Quick Actions</h1>
...
+
+        const continueButtonElement = document.getElementById(
+            'continueButton'
+        );
+
+        const quickActionStatusTable = document.getElementById(
+            'quickActionStatusTable'
+        );
+
+        generateQuickActionsButtonElement.addEventListener('click', () => {
```
khawkins, line 85: You have two of these, the other on line 93 below.

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -56,54 +56,87 @@ <h1>Create sObject LWC Quick Actions</h1>
...
+            webviewMessaging.sendMessageRequest(
+                'generateLwcQuickActionsButton',
```
khawkins, line 87: Nit: these requests might benefit from more descriptive names of the _action_ requested, rather than the _element_ initiating the action. Same for the `continueButtonElement` click event below.

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -1,3 +1,10 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
```
khawkins, line 2: Haha thank you. 😆

src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -27,10 +35,53 @@ export type GetSObjectsStatus = {
...
+                    {
+                        type: 'continueButton',
+                        action: (panel) => {
+                            panel.dispose();
+                            return resolve();
+                        }
+                    },
+                    {
+                        type: 'generateLwcQuickActionsButton',
+                        action: async (panel, _data, callback) => {
```
khawkins, line 55: Nit: may as well `_panel` too, as it's not used either.

src/utils/codeBuilder.ts
```diff
@@ -0,0 +1,309 @@
...
+import { Uri } from 'vscode';
+import * as path from 'path';
+import { CompactLayoutField } from './orgUtils';
+
+type TemplateVariables = { [name: string]: string };
+
+export class CodeBuilder {
+    static readonly TEMPLATE_DIR = './resources/templates';
+    static readonly QUICK_ACTION_TEMPLATE_NAME = 'quickAction.xml';
+    static readonly LWC_DESTINATION_DIR = './force-app/main/default/lwc';
```
khawkins, line 17: I wonder if we should standardize on the familiar SFDX project path constants living in `workspaceUtils.ts`, similar to how we did for [`STATIC_RESOURCES_PATH`](https://github.com/salesforce/salesforcedx-vscode-mobile/blob/7632f668981bd3938045f0d7d3e6be6054dd1980/src/utils/workspaceUtils.ts#L13-L18)? I ran into issues where I originally started defining those paths in one of the command files, and then ended up needing them in different commands and utilities.

src/utils/codeBuilder.ts
```diff
@@ -0,0 +1,309 @@
...
+            });
+        });
+        return fieldNames;
+    }
+
+    private createQuickAction(
+        templateVariables: TemplateVariables,
+        label: string,
+        name: string,
+        iconName: string | undefined = undefined
```
khawkins, line 122: Nit: optional properties will default to `undefined`, if they're not passed in. You don't need the default assignment.

src/utils/codeBuilder.ts
```diff
@@ -0,0 +1,309 @@
...
+
+    private writeFileContents(
+        dirPath: string,
+        filename: string,
+        content: string
+    ) {
+        // ensure dirPath exists
+        if (!fs.existsSync(dirPath)) {
+            try {
+                fs.mkdirSync(dirPath, { recursive: true });
```
khawkins, line 229: I've been a little IDK on this approach. Ideally we're operating from a place where the user has properly bootstrapped their project folder to be synced with the Offline Starter Kit project—where all of these folders should already exist. That it might not exist points to a more fundamental issue to me, than this code not functioning. But then you have to do some sanity checking beforehand, falling through on failure there, etc. ¯\\_(ツ)_/¯ not sure of the right way to handle those boundary checks.

khawkins: I.e. I'm not sure if we shouldn't be making more liberal use of the [`ConfigureProjectCommand.validateProjectFolder()` method](https://github.com/salesforce/salesforcedx-vscode-mobile/blob/7632f668981bd3938045f0d7d3e6be6054dd1980/src/commands/wizard/configureProjectCommand.ts#L247-L310) in the up-front validations of these commands that depend upon a functional project folder. Might be food for thought, for another day.

resources/instructions/createSObjectLwcQuickActions.html
dbreese: Opps, that was a merge conflict issue!

resources/instructions/createSObjectLwcQuickActions.html
dbreese: Good idea, will rename

src/utils/codeBuilder.ts
```diff
@@ -0,0 +1,309 @@
...
+        var imports = '';
+        var createFieldsHtml = '';
+        var editFieldsHtml = '';
+        var importAliases = '';
+        var variableAssignments = '';
+
+        fieldNames.forEach((field) => {
+            var fieldNameImport = `${field.toUpperCase()}_FIELD`;
+            fields += `${fieldNameImport}, `;
+            imports += `import ${fieldNameImport} from "@salesforce/schema/${this.objectApiName}.${field}";\n`;
```
khawkins, line 287: Do you think it would be worthwhile to represent these template strings as constants at the top of the file? So that you'd have something like:

```typescript
const IMPORT_STATEMENT_TEMPLATE = 'import <FieldNameImport> from "@salesforce/schema/<ObjectApiName>.<FieldName>";'
const LIGHTNING_INPUT_FIELD_TEMPLATE = '<lightning-input-field field-name={<FieldNameVariable>} value={<FieldName>}></lightning-input-field>';
```
And then do substitutions here? I can't decide if that would make things more readable or less readable.

src/utils/codeBuilder.ts
dbreese: I think that is a good idea. I will move these to that location.

src/utils/codeBuilder.ts
dbreese: I am getting a compile error if I dont specify it from the invocations above.

src/utils/codeBuilder.ts
dbreese: Yeah, I'm not sure either. I am not a fan of all this string replacement, but not sure what else to do. Since it is hard to read already, moving the strings someplace else I almost feel like it could be more difficult to debug and/or maintain. But, I'm up for whatever!

dbreese:
Looks like it is failing for Lead LWC and QA generation. Debugging now. I also had to reboot vscode and clear node-modules and clean to finally get app to run again, so it could be related to that.

dbreese:
Ahhhh, it is not finding Visit because I dont have the licenses set up in my org...crappy edge case that will probably bite a lot of people. Hmmm.....

src/utils/codeBuilder.ts
khawkins: Ah, it's because of the specific nomenclature, which I always get wrong: for optional method parameters specifically, you have to declare the optional input parameter as `iconName?: string`. This is distinct from declaring a variable as type `<type> | undefined` (e.g. `let myOptionalString: string | undefined;`. The fact that these feel like a distinction without a difference is what always makes me forget which one to use where.

Anyway, long story short: the proper way to code this optional parameter in a method, in TypeScript, is to declare it as `iconName?: string`.

src/utils/codeBuilder.ts
khawkins: > moving the strings someplace else I almost feel like it could be more difficult to debug and/or maintain

That was my concern too. I'm okay if we leave it, unless I think of some far more compelling way to make it more readable—which I haven't come up with yet. 😛

src/utils/codeBuilder.ts
dbreese: Aha! Updated!

src/utils/codeBuilder.ts
```diff
@@ -0,0 +1,307 @@
...
+            });
+        });
+        return fieldNames;
+    }
+
+    private createQuickAction(
+        templateVariables: TemplateVariables,
+        label: string,
+        name: string,
+        iconName?: string | undefined
```
khawkins, line 120: Remaining nit: you don't need the `| undefined`. The `?:` implies that it's either the specified type, or not provided—which equates to `undefined`.

khawkins (approved):
Looks good overall!

src/utils/codeBuilder.ts
dbreese: Ahhhhh, sorry! Will update and push!

PR #53, opened 2023-11-10
Get correct compact layout for sObject

sfdctaka:
To get a compact layout for a sObject I am using 2-pass process.

If sObject has compact layout assigned it will have `defaultCompactLayoutId` assigned. Otherwise, `null` is assigned and in that case the compact layout is defaults to `SYSTEM` compact layout.

After determining which compact layout is actually needed only after that a second network request need to be made in order to get the actual layout content.

### DIFF
--- a/src/utils/orgUtils.ts
+++ b/src/utils/orgUtils.ts
@@ -72,13 +72,13 @@ export class OrgUtils {
         }
     }
 
-    public static async getAllCompactLayoutsForSObject(
+    public static async getCompactLayoutsForSObject(
         sObjectName: string
-    ): Promise<any> {
+    ): Promise<Object> {
         const org = await Org.create();
         const conn = org.getConnection();
 
-        const result = await conn.request(
+        const result = await conn.request<Object>(
             `/services/data/v59.0/sobjects/${sObjectName}/describe/compactLayouts`
         );
 
@@ -88,11 +88,11 @@ export class OrgUtils {
     public static async getCompactLayoutForSObject(
         sObjectName: string,
         recordTypeId: string
-    ): Promise<any> {
+    ): Promise<Object> {
         const org = await Org.create();
         const conn = org.getConnection();
 
-        const result = await conn.request(
+        const result = await conn.request<Object>(
             `/services/data/v59.0/sobjects/${sObjectName}/describe/compactLayouts/${recordTypeId}`
         );
 
@@ -106,21 +106,16 @@ export class OrgUtils {
             const fields: CompactLayoutField[] = [];
 
             // Get all the compact layouts associated to this sObject first
-            let result = await this.getAllCompactLayoutsForSObject(sObjectName);
+            let result = await this.getCompactLayoutsForSObject(sObjectName);
 
             if (result) {
-                const resultObj = result as Object;
-
                 // sObject can have multiple compact layouts associated with it. Get the default.
                 const defaultCompactLayoutId =
-                    resultObj['defaultCompactLayoutId' as keyof Object];
+                    result['defaultCompactLayoutId' as keyof Object];
 
                 // Mapping tab
-                console.error(resultObj);
                 const recordTypeCompactLayoutMappings =
-                    resultObj[
-                        'recordTypeCompactLayoutMappings' as keyof Object
-                    ];
+                    result['recordTypeCompactLayoutMappings' as keyof Object];
 
                 // ID of compact layout need to be normalized
                 const recordTypeCompactLayoutMapping = (
--- a/src/utils/orgUtils.ts
+++ b/src/utils/orgUtils.ts
@@ -36,6 +36,10 @@ export type SObjectCompactLayouts = {
     recordTypeCompactLayoutMappings: SObjectCompactLayoutMapping[];
 };
 
+export type SObjectCompactLayout = {
+    fieldItems: CompactLayoutField[];
+};
+
 export class OrgUtils {
     public static async getSobjects(): Promise<SObject[]> {
         try {
@@ -99,11 +103,11 @@ export class OrgUtils {
     public static async getCompactLayoutForSObject(
         sObjectName: string,
         recordTypeId: string
-    ): Promise<Object> {
+    ): Promise<SObjectCompactLayout> {
         const org = await Org.create();
         const conn = org.getConnection();
 
-        const result = await conn.request<Object>(
+        const result = await conn.request<SObjectCompactLayout>(
             `/services/data/v59.0/sobjects/${sObjectName}/describe/compactLayouts/${recordTypeId}`
         );
 
@@ -124,9 +128,8 @@ export class OrgUtils {
                 const defaultCompactLayoutId = result.defaultCompactLayoutId;
                 result['defaultCompactLayoutId'];
 
-                // Mapping tab
-                const recordTypeCompactLayoutMappings =
-                    result.recordTypeCompactLayoutMappings;
+                // Mapping table
+                const recordTypeCompactLayoutMappings = result.recordTypeCompactLayoutMappings;
 
                 // ID of compact layout need to be normalized
                 const recordTypeCompactLayoutMapping =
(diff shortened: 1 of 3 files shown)


src/utils/orgUtils.ts
```diff
@@ -18,6 +18,13 @@ export interface Field {
     label: string;
     type: string;
 }
+
+export interface CompactLayoutField {
+    editableForNew: boolean;
+    editableForUpdate: boolean;
+    label: string;
+}
```
sfdctaka, line 26: @dbreese We don't have to use this interface. I added so that I can cast to this before returning to the caller so LWC generating code can have defined fields in the elements of returned array of sObject's fields. Otherwise, you might need to do lot of casting as I do in `getCompactLayoutFieldsForSObject`.

src/utils/orgUtils.ts
```diff
@@ -65,6 +72,103 @@ export class OrgUtils {
...
+    ): Promise<any> {
+        const org = await Org.create();
+        const conn = org.getConnection();
+
+        const result = await conn.request(
+            `/services/data/v59.0/sobjects/${sObjectName}/describe/compactLayouts/${recordTypeId}`
+        );
+
+        return Promise.resolve(result);
+    }
```
sfdctaka, line 100: I'm wrapping `conn.request` because `request` method of `Connection` returns `StreamResult`. `StreamResult` is a type from `@salesforce/core` API and I couldn't find a way to create it or use it. Worse, Typescript was very strict when I tried to create stubs using sinon. So I wrapped them in methods and this way I can wrap the wrapping methods in the unit tests.

src/utils/orgUtils.ts
```diff
@@ -65,6 +72,103 @@ export class OrgUtils {
...
+        sObjectName: string
+    ): Promise<CompactLayoutField[]> {
+        try {
+            const fields: CompactLayoutField[] = [];
+
+            // Get all the compact layouts associated to this sObject first
+            let result = await this.getAllCompactLayoutsForSObject(sObjectName);
+
+            if (result) {
+                const resultObj = result as Object;
```
khawkins, line 112: If we're assuming any value for `result` is an `Object`, can `getAllCompactLayoutsForSObject()` return `Promise<Object>` instead of `Promise<any>`?

src/utils/orgUtils.ts
khawkins: It looks like it returns a `StreamPromise` from some version of the `jsforce` package—which is just an extension of `Promise` that has some stream-specific functions added.

But I think it's fine you're doing it this way, in any case. As I mentioned below, I would change the type to `Promise<Object>`, and coerce that type as necessary.

If we wanted to get fancy at some point, the `request` method takes a template type, so we could be more definitive in specifying what we want to return, for example:

```typescript
const result: Promise<OurStronglyTypedCompactLayoutResultType> =
    await conn.request<OurStronglyTypedCompactLayoutResultType>(
            `/services/data/v59.0/sobjects/${sObjectName}/describe/compactLayouts/${recordTypeId}`
    );
return Promise.resolve(result);
```

src/utils/orgUtils.ts
```diff
@@ -65,6 +72,103 @@ export class OrgUtils {
         }
     }
 
+    public static async getAllCompactLayoutsForSObject(
```
khawkins, line 75: Nit: I'd just call this `getCompactLayoutsForSObject`. The "all" is pretty much implied.

src/utils/orgUtils.ts
```diff
@@ -65,6 +72,98 @@ export class OrgUtils {
...
+            let result = await this.getCompactLayoutsForSObject(sObjectName);
+
+            if (result) {
+                // sObject can have multiple compact layouts associated with it. Get the default.
+                const defaultCompactLayoutId =
+                    result['defaultCompactLayoutId' as keyof Object];
+
+                // Mapping tab
+                const recordTypeCompactLayoutMappings =
+                    result['recordTypeCompactLayoutMappings' as keyof Object];
```
khawkins, line 118: Nit: `as keyof Object` here doesn't really offer anything, and is actually somewhat misleading. `as keyof <SomeType>` is meant to offer a type assertion that the property name falls under the explicit list of properties defined by `SomeType`. It's basically another form of a type check on the named property you're giving in the square bracket notation.

In this case, `keyof Object` doesn't really make sense, because `Object` doesn't specify any of these property names in its definition—they're ours. It seems to be happily working anyway, but it's misleading nomenclature in this case. See the [`keyof` docs](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html) for more information.

src/utils/orgUtils.ts
```diff
@@ -65,6 +72,98 @@ export class OrgUtils {
...
+
+                // ID of compact layout need to be normalized
+                const recordTypeCompactLayoutMapping = (
+                    recordTypeCompactLayoutMappings as unknown as Array<Object>
```
khawkins, line 122: What does the intermediary coercion to `unknown` do here? Is that required?

src/utils/orgUtils.ts
sfdctaka: Without `as keyof Object` this is what I get in the editor.
image

src/utils/orgUtils.ts
sfdctaka: Similar to above without a hard cast Typescript would complain.
image

khawkins: Oh I get the issue now. In this case, I'm going to recommend my previous pie-in-the-sky type coercion be done now. The code gets hard to read all over the place ,because we're basically hedging against defining the types we expect to work with in the response, in favor of "cast to any" statements all over the place, which in turn is punting on type safety.

Let's just create the types instead. From the code you've got, I can glean the following type definitions that represent the data we expect to get in these results:

```typescript
export type SObjectCompactLayoutMapping = {
    compactLayoutId: string;
    compactLayoutName: string;
    recordTypeId: string;
};

export type SObjectCompactLayouts = {
    defaultCompactLayoutId?: string;
    recordTypeCompactLayoutMappings: SObjectCompactLayoutMapping[];
    fieldItems: CompactLayoutField[];
};
```

I mean, we can tweak these, change type names, etc. But if we tack to an approach of returning `Promise<SObjectCompactLayouts>` instead of `Object`, we can access all of the expected properties as first-class citizens rather than casts to generic types.

dbreese (approved):

src/utils/orgUtils.ts
```diff
@@ -65,6 +83,97 @@ export class OrgUtils {
...
+                            return (
+                                element.compactLayoutId ===
+                                defaultCompactLayoutId
+                            );
+                        } else {
+                            // defaultCompactLayoutId can be null when a compact layout is not assigned.
+                            // In that case sObject will always have one default layout called SYSTEM.
+                            // So use that instead.
+                            const compactLayoutName = element[
+                                'compactLayoutName' as keyof Object
```
khawkins, line 144: Can we update all of these, based on the previous feedback? We should be able avoid coercing a key type and having to cast properties to `unknown`.

- `compactLayoutName` is already in the [`SObjectCompactLayoutMapping` type](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/53/files#diff-9059df4604057fed56162133c1646eececc243a7f261ac69ffb90d57918c0f70R28-R31).
- `recordTypeId`, used [below](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/53/files#diff-9059df4604057fed56162133c1646eececc243a7f261ac69ffb90d57918c0f70R151-R154), is too.
- My read of what comes back from `getCompactLayoutForSObject()` is:
    - An `SObjectCompactLayoutMapping` instance, and then that type requires:
    - An additional `fieldItems` property of type `CompactLayoutField[]`, based on [the code below](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/53/files#diff-9059df4604057fed56162133c1646eececc243a7f261ac69ffb90d57918c0f70R164-R166).  
    
    We should update `getCompactLayoutForSObject()` to return `Promise<SObjectCompactLayoutMapping>` accordingly, and that should close the loop.

sfdctaka: Updated!

khawkins (approved):
Thanks @sfdctaka ! This is looking like a great first pass! 👍

PR #61, opened 2024-02-20
Handle case where an error occurs of any type when attempting to get page layout for missing objects.

dbreese:
Screenshot 2024-02-20 at 12 19 08 PM

@khawkins @maliroteh-sf @sfdctaka

### DIFF
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -55,7 +55,7 @@
             </tbody>
         </table>
 
-        <div id="errorDiv">
+        <div id="errorDiv" style="color: crimson">
             <!-- placeholder for any error messages -->
         </div>
 
--- a/resources/instructions/createSObjectLwcQuickActions.html
+++ b/resources/instructions/createSObjectLwcQuickActions.html
@@ -54,6 +54,10 @@
                 <!-- will be filled in programatically below -->
             </tbody>
         </table>
+        
+        <div id="errorDiv">
+            <!-- placeholder for any error messages -->
+        </div>
 
         <label
             style="display: block; margin-top: 10px; margin-bottom: 8px"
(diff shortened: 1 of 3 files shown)


src/commands/wizard/lwcGenerationCommand.ts
```diff
@@ -101,107 +116,117 @@ export class LwcGenerationCommand {
...
             }
-
-            resolve(getSObjectsStatus);
         });
     }
 
     static async checkForExistingQuickActions(): Promise<SObjectQuickActionStatus> {
-        return new Promise<SObjectQuickActionStatus>(async (resolve) => {
-            const results: SObjectQuickActionStatus = { sobjects: {} };
+        return new Promise<SObjectQuickActionStatus>(
```
sfdctaka, line 130: It looks like only the formatting changed. Was it by running prettier?

sfdctaka (approved):
LGTM!

resources/instructions/createSObjectLwcQuickActions.html
```diff
@@ -55,6 +55,10 @@ <h1>Create sObject LWC Quick Actions</h1>
             </tbody>
         </table>
 
+        <div id="errorDiv">
```
maliroteh-sf, line 58: nit: It would be nice to set a text styling here for the error messages so as to grab user's attention. We could start with something simple like `<div id="errorDiv" style="color:crimson;">` and let UX fine tune it later.

maliroteh-sf (approved):

src/commands/wizard/lwcGenerationCommand.ts
dbreese: This file changed mostly because of prettier, but also because the lines grew in length some because I added the `reject` block to the Promise.

resources/instructions/createSObjectLwcQuickActions.html
dbreese: Good idea, haven't received designs from product or UX, so just left it pretty bland, but I like your idea. Here it is:

Screenshot 2024-02-20 at 1 30 21 PM

PR #62, opened 2024-02-29
Add a command to install eslint plugin

sfdctaka:
This command will add an item in `devDependencies` of `package.json`. That item is `eslint-plugin-lwc-graph-analyzer`.

If there is no `.eslintrc.json` or is missing to add configuration there that would be added too.


package.json
```diff
@@ -40,6 +40,12 @@
         "command": "salesforcedx-vscode-offline-app.onboardingWizard",
         "title": "%extension.commands.config-wizard.title%",
         "category": "%extension.commands.config-wizard.category%"
+      },
+      {
+        "command": "salesforcedx-vscode-offline-app.configureLintingTools",
+        "title": "%extension.commands.config-linting-tools.title%",
+        "category": "%extension.commands.config-wizard.category%",
+        "when": "sfdx_project_opened"
```
sfdctaka, line 48: @maliroteh-sf I can't get `when` clause to not show the command if the folder opened in VSCode is not SFDX project. Do you know what I may be doing incorrectly?

sfdctaka:
@maliroteh-sf @dbreese @khawkins May I get help from you on the text copies please?

package.json
maliroteh-sf: That's b/c you must define the `menus` section too. So the `contributes` section should look like this instead:
```
"contributes": {
  "menus": {
    "commandPalette": [
      {
        "command": "salesforcedx-vscode-offline-app.configureLintingTools",
        "when": "sfdx_project_opened"
      }
    ]
  },
  "commands": [
    {
      "command": "salesforcedx-vscode-offline-app.onboardingWizard",
      "title": "%extension.commands.config-wizard.title%",
      "category": "%extension.commands.config-wizard.category%"
    },
    {
      "command": "salesforcedx-vscode-offline-app.configureLintingTools",
      "title": "%extension.commands.config-linting-tools.title%",
      "category": "%extension.commands.config-wizard.category%"
    }
  ]
},
```

src/utils/workspaceUtils.ts
```diff
@@ -67,6 +69,38 @@ export class WorkspaceUtils {
...
+
+    static setPackageJson(packageJson: object) {
+        fs.writeFileSync(
+            path.join(this.getWorkspaceDir(), PACKAGE_JSON),
+            JSON.stringify(packageJson, null, TAB_SPACES)
+        );
+    }
+
+    static packageJsonExists(): boolean {
+        return fs.existsSync(path.join(this.getWorkspaceDir(), PACKAGE_JSON));
```
maliroteh-sf, line 90: See my comment below for `isSfdxProjectOpened`.... I think `packageJsonExists()` should never throw and instead should always return either `true` or `false` Since  `getWorkspaceDir()` can throw you should wrap this code in `try-catch` to ensure you always return either `true` or `false`

maliroteh-sf (changes requested):

src/utils/workspaceUtils.ts
```diff
@@ -67,6 +69,38 @@ export class WorkspaceUtils {
...
+    }
+
+    static lwcFolderExists(): boolean {
+        return fs.existsSync(
+            path.join(this.getWorkspaceDir(), WorkspaceUtils.LWC_PATH)
+        );
+    }
+
+    static isSfdxProjectOpened(): boolean {
```
maliroteh-sf, line 99: This is not correct. You should also add [this line](https://git.soma.salesforce.com/MobilePlatform/salesforcedx-vscode-mobile/blob/browser-container/src/extension.ts#L58) from the code that I shared with you.

Basically `isSfdxProjectOpened()` should never throw and instead should return either `true` or `false` but as you have defined it right now, `getWorkspaceDir()` can throw. So either at the line above or wrap the whole thing in `try-catch`

```
static isSfdxProjectOpened(): boolean {
    try {
        return fs.existsSync(
            path.join(this.getWorkspaceDir(), SFDX_PROJECT_FILE)
        );
    } catch {
        return false;
    }
}
```

src/utils/workspaceUtils.ts
sfdctaka: Does `fs.existsSync()` throw error? The return value is boolean so if an error is thrown internally in that method it should be caught and return `false`. So basically it should be what you suggested inside it. Let me check, that's what I thought I read it somewhere.

sfdctaka: https://github.com/nodejs/node/blob/40565e90b4fc72649cd4801b01d6e0e00f21b06c/lib/fs.js#L283

sfdctaka: Nvm. You meant getWorkspaceDir can throw. I'm going to fix it. Thanks.

package.json
```diff
@@ -35,11 +35,25 @@
...
       {
         "command": "salesforcedx-vscode-offline-app.onboardingWizard",
         "title": "%extension.commands.config-wizard.title%",
         "category": "%extension.commands.config-wizard.category%"
+      },
+      {
+        "command": "salesforcedx-vscode-offline-app.configureLintingTools",
+        "title": "%extension.commands.config-linting-tools.title%",
+        "category": "%extension.commands.config-wizard.category%",
+        "when": "sfdx_project_opened"
```
maliroteh-sf, line 56: This should not be here anymore. The `when` clause should not be under `commands` but under `menus`

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,202 @@
...
+    InformationYesNo,
+    InformationOk
+}
+
+export function onActivate(context: ExtensionContext) {
+    commands.executeCommand(
+        'setContext',
+        'sfdx_project_opened',
+        WorkspaceUtils.isSfdxProjectOpened()
+    );
```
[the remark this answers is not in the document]
maliroteh-sf: Yes I'm sure. It works right now b/c you only have `sfdx_project_opened` called in `onActivate` of this command only.... but that is bad design. You are setting the context for the entire extension but inside the `onActivate` of this command. The correct design is to do it inside the `onActivate` of the extension itself. Then the `onActivate` of this command can be deleted altogether.

maliroteh-sf (approved):
👍

package.json
```diff
@@ -35,11 +35,24 @@
   "main": "./out/extension.js",
   "l10n": "./l10n",
   "contributes": {
+    "menus": {
+      "commandPalette": [
+        {
+          "command": "salesforcedx-vscode-offline-app.configureLintingTools",
+          "when": "sfdx_project_opened"
```
khawkins, line 42: Oh cool, I didn't know about this functionality. That's a clean way to specify a precondition here.

package.json
```diff
@@ -35,11 +35,24 @@
...
     "commands": [
       {
         "command": "salesforcedx-vscode-offline-app.onboardingWizard",
         "title": "%extension.commands.config-wizard.title%",
         "category": "%extension.commands.config-wizard.category%"
+      },
+      {
+        "command": "salesforcedx-vscode-offline-app.configureLintingTools",
+        "title": "%extension.commands.config-linting-tools.title%",
+        "category": "%extension.commands.config-wizard.category%"
```
khawkins, line 55: Hmmm, I'm not sure about this. In theory this doesn't just apply to the Starter Kit—or really at all, because the Starter Kit already has the linting rules configured out of the box. I wonder if we should define a new category, perhaps called "Salesforce Mobile Offline"?

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+import { commands, l10n, window, ExtensionContext } from 'vscode';
+import * as fs from 'fs';
+import * as path from 'path';
+import { WorkspaceUtils } from '../../utils/workspaceUtils';
+import { TAB_SPACES } from '../../utils/constants';
+
+const configureLintingToolsCommand =
+    'salesforcedx-vscode-offline-app.configureLintingTools';
+const eslintDependencies = [
+    ['@salesforce/eslint-plugin-lwc-graph-analyzer', '^0.9.0'],
```
khawkins, line 17: These are configuration items where at least the version values would be better represented in [contributes.configuration](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) settings. That way a) we take hard-coded data out of the code, and b) users can update these values in their own personal configuration settings, to override the versions we select.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+                    title: l10n.t('OK')
+                });
+            case MessageType.InformationYesNo:
+                return await window.showInformationMessage(
+                    localizedMessage,
+                    { title: l10n.t('Yes') },
+                    { title: l10n.t('No') }
+                );
+            case MessageType.InformationOk:
+                return await await window.showInformationMessage(
```
khawkins, line 180: You have one too many `await`s here.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+    Error,
+    InformationYesNo,
+    InformationOk
+}
+
+export class ConfigureLintingToolsCommand {
+    static async configure(): Promise<boolean> {
+        try {
+            if (!WorkspaceUtils.lwcFolderExists()) {
+                await this.showMessage('LWC folder does not exist.');
```
khawkins, line 38: Do these messages need any further context? I don't know what the message title/header will look like, if any. But on their own, these messages seem a little sparse/ambiguous.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+            if (!WorkspaceUtils.packageJsonExists()) {
+                await this.showMessage(
+                    'The project does not contain package.json.'
+                );
+                return Promise.resolve(false);
+            }
+
+            // Ask user to add eslint plugin
+            const result = await this.showMessage(
+                'Do you want to add eslint plugin for LWC data graph anaylsis to your package.json?',
```
khawkins, line 51:
```suggestion
                'Do you want to add the ESLint plugin for LWC graph analysis to your project? This will give you linting feedback on code patterns that will not support your LWCs working offline, for mobile use cases.',
```

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+            WorkspaceUtils.getWorkspaceDir(),
+            WorkspaceUtils.LWC_PATH,
+            '.eslintrc.json'
+        );
+
+        if (fs.existsSync(eslintrcPath)) {
+            const eslintrc = JSON.parse(fs.readFileSync(eslintrcPath, 'utf-8'));
+            const eslintrcExtends = eslintrc.extends as Array<string>;
+
+            if (eslintrc && eslintrcExtends) {
```
khawkins, line 133: Should we not add the `extends` item, if it doesn't exist?

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+                        eslintrcPath,
+                        JSON.stringify(eslintrc, null, TAB_SPACES)
+                    );
+                }
+            }
+        } else {
+            // Create eslintrc
+            fs.writeFileSync(
+                eslintrcPath,
+                `{"extends": ["${eslintRecommended}", "${lwcGraphAnalyzerRecommended}"]}`
```
khawkins, line 158: To be consistent, it feels like we should build this up and save in the same format as we do for existing files. I.e. make a JSON object, and stringify it with `TAB_SPACES` to write it to disk.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+                    modified = this.updateDevDependencies();
+                } catch (error) {
+                    await this.showMessage(
+                        `Error updating package.json: ${error}`
+                    );
+                    return Promise.resolve(false);
+                }
+
+                try {
+                    this.updateEslintrc();
```
khawkins, line 69: We should be tracking whether this file was updated or not too, and reporting to the user accordingly—even if the report is built into a single message. We could be updating this but not the `devDependencies`, and vice versa. We should be clear to the user, exactly what's being updated and what is not.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+                } catch (error) {
+                    await this.showMessage(
+                        `Error updating .eslintrc.json: ${error}`
+                    );
+                    return Promise.resolve(false);
+                }
+
+                if (modified) {
+                    await this.showMessage(
+                        `Updated developer dependency in package.json. Run package manager such as npmr/yarn/pnpm to update node modules.`,
```
khawkins, line 79:
```suggestion
                        `Updated package.json to include offline linting package dependencies. In the Terminal window, be sure to run the install command for your configured package manager, to install the updated dependencies. For example, "npm install" or "yarn install".`,
```

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+                        MessageType.InformationOk
+                    );
+                } else {
+                    await this.showMessage(
+                        `No update was made in package.json. It already includes eslint plugin for LWC data graph analysis.`,
```
khawkins, line 84:
```suggestion
                        `All offline linting packages and dependencies are already configured in your project. No update has been made to package.json.`,
```

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,194 @@
...
+                        MessageType.InformationOk
+                    );
+                }
+
+                return Promise.resolve(true);
+            }
+        } catch (error) {
+            await this.showMessage(
+                `There was an error trying to update developer dependency in package.json: ${error}`
```
khawkins, line 93:
```suggestion
                `There was an error trying to update either the offline linting dependencies or linting configuration: ${error}`
```

src/utils/constants.ts
```diff
@@ -7,3 +7,6 @@
 
 export const MINIMUM_REQUIRED_VERSION_CORE_EXTENSION = '58.4.1';
 export const CORE_EXTENSION_ID = 'salesforce.salesforcedx-vscode-core';
+export const SFDX_PROJECT_FILE = 'sfdx-project.json';
+export const PACKAGE_JSON = 'package.json';
+export const TAB_SPACES = 2;
```
khawkins, line 12: Nit: Can we name this a bit more descriptively, e.g. `JSON_INDENTATION_SPACES`?

src/commands/lint/configureLintingToolsCommand.ts
sfdctaka: Showing 3 messages now at a maximum, one for updating `devdependencies`, one for updating `.eslintrc.json`, and one for running npm. I thought maybe I should mash the three. But localizing string was going to be tricky so instead went with three.

image

src/commands/lint/configureLintingToolsCommand.ts
sfdctaka: image

package.json
```diff
@@ -35,13 +35,41 @@
...
+      {
+        "command": "salesforcedx-vscode-offline-app.configureLintingTools",
+        "title": "%extension.commands.config-linting-tools.title%",
+        "category": "%extension.commands.salesforce-mobile-offline.category%"
+      }
+    ],
+    "configuration": {
+      "title": "%salesforce.mobile.extensions%",
+      "properties": {
+        "mobileOfflineLinting.eslint-plugin-lwc-graph-analyzer": {
```
sfdctaka, line 61: Without preceding `mobileOfflineLinting` separated by a dot the configuration doesn't get rendered in the settings page of VSCode. With it it becomes category name separated by a colon.

image

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,241 @@
...
+    InformationYesNo,
+    InformationOk
+}
+
+export class ConfigureLintingToolsCommand {
+    static async configure(): Promise<boolean> {
+        try {
+            if (!WorkspaceUtils.lwcFolderExists()) {
+                await this.showMessage(
+                    'LWC folder does not exist. The folder is required to create a configuration file for eslint.'
```
khawkins, line 54:
```suggestion
                    '"force-app/main/default/lwc" folder does not exist. This folder is required to create a configuration file for ESLint.'
```

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -0,0 +1,241 @@
...
+                );
+                return Promise.resolve(false);
+            }
+
+            if (!WorkspaceUtils.packageJsonExists()) {
+                await this.showMessage(
+                    'The sfdx project does not contain package.json. It is required to specify dependent eslint packages.'
```
khawkins, line 61:
```suggestion
                    'Your project does not contain a "package.json" specification. You must have a package specification to configure these ESLint packages and their dependencies in your project.'
```

PR #66, opened 2024-06-21
fix: update sfdx command to sf to match dependent command id change

ben-zhang-at-salesforce:
salesforcedx-vscode-core changed all sfdx.* command to sf.*. we need to match the change

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,7 @@
         "out": true // set this to false to include "out" folder in search results
     },
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
-    "typescript.tsc.autoDetect": "off"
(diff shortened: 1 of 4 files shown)


maliroteh-sf (commented):
Need to fix prettier issues so that CI jobs all pass.

khawkins (approved):
Looks like this needs `npm run prettier:write` changes committed, but other than that LGTM!

ben-zhang-at-salesforce:
now this pr is failing with same error as https://github.com/salesforce/salesforcedx-vscode-mobile/pull/67

maliroteh-sf:
> now this pr is failing with same error as #67

@ben-zhang-at-salesforce wait for the fix in #67 to go in, then rebase and you should be good to go.

PR #67, opened 2024-06-24
Telling VSCode that Prettier is the default formatter

khawkins:
Just helps keep VSCode formatting synced up with what we align to in our Node project.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,7 @@
         "out": true // set this to false to include "out" folder in search results
     },
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
-    "typescript.tsc.autoDetect": "off"
+    "typescript.tsc.autoDetect": "off",
+    // Prettier is this project's code formatter.
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
 }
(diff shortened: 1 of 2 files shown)


maliroteh-sf (approved):

khawkins:
The test failures are a bit of an unwinding, here. I went looking for the cause of this error:

> Failed to run tests Error: spawnSync D:\a\salesforcedx-vscode-mobile\salesforcedx-vscode-mobile\.vscode-test\vscode-win32-x64-archive-1.90.2\bin\code.cmd EINVAL

which looked to be fixed by https://github.com/microsoft/vscode-test-cli/pull/46, except `vscode-test-cli` is new news to me. Went to [Microsoft's docs](https://code.visualstudio.com/api/working-with-extensions/testing-extension) on testing your extension, and it looks like they've quietly rolled out a new Node package to assist in the previous setup.

Going to see what it takes to change our test runner accordingly.

khawkins:
NB: Didn't need to change our test runner over to the other package—and couldn't anyway, because (AFAICT) the need to install the Salesforce Extensions pushes us into the "advanced" setup use case of calling `test-electron` functions a little more manually.

I ported the changes from the other upstream projects, to our test runner. And tests are working again. @maliroteh-sf I don't know if you want to take another look or not, at those changes specifically.

maliroteh-sf (approved):

PR #68, opened 2024-06-25
feat: config lwc mobile eslint plugin

ben-zhang-at-salesforce:
Scaffolding up adding lwc-mobile eslint plugin into salesforcedx-vscode-mobile.  

not fully read yet, need to wait for lwc-mobile eslint plugin release to npmjs.com

### DIFF
--- a/src/commands/lint/configureLintingToolsCommand.ts
+++ b/src/commands/lint/configureLintingToolsCommand.ts
@@ -18,19 +18,21 @@ const config = workspace.getConfiguration();
 
 class EslintDependencyConfig {
     readonly name: string;
-    readonly packageJsonPropertyId: string;
-    readonly version: string;
-    readonly eslintrcExtend: string;
+    readonly packageConfigPropertyId: string;
+    readonly eslintConfigToExtend: string;
 
     constructor(
         name: string,
-        packageJsonPropertyId: string,
-        eslintrcExtend: string
+        packageConfigPropertyId: string,
+        eslintConfigToExtend: string
     ) {
         this.name = name;
-        this.packageJsonPropertyId = packageJsonPropertyId;
-        this.eslintrcExtend = eslintrcExtend;
-        this.version = config.get(this.packageJsonPropertyId) as string;
+        this.packageConfigPropertyId = packageConfigPropertyId;
+        this.eslintConfigToExtend = eslintConfigToExtend;
+    }
+
+    getVersion(): string {
+        return config.get(this.packageConfigPropertyId) as string;
     }
 }
 
@@ -81,7 +83,7 @@ export class ConfigureLintingToolsCommand {
 
             // Ask user to add eslint plugin
             const result = await this.showMessage(
-                'Do you want to add the ESLint plugins related to LWC offline capabilities to your project? This will give you linting feedback on code patterns that will not support your LWCs working offline, for mobile use cases.',
+                'Do you want to add Salesforce code linting guidance for Mobile and Offline capabilities? These tools will identify code patterns that cause problems in Mobile and Offline use cases.',
                 MessageType.InformationYesNo
             );
 
@@ -153,9 +155,9 @@ export class ConfigureLintingToolsCommand {
 
         if (devDependencies) {
             eslintDependencies.forEach((dependencyConfig) => {
-                const { name, version } = dependencyConfig;
+                const { name } = dependencyConfig;
                 if (!devDependencies[name]) {
-                    devDependencies[name] = version;
+                    devDependencies[name] = dependencyConfig.getVersion();
                     modified = true;
                 }
             });
@@ -188,8 +190,8 @@ export class ConfigureLintingToolsCommand {
             let modified = false;
 
             eslintDependencies.forEach((config) => {
-                if (!eslintrcExtends.includes(config.eslintrcExtend)) {
-                    eslintrcExtends.push(config.eslintrcExtend);
+                if (!eslintrcExtends.includes(config.eslintConfigToExtend)) {
+                    eslintrcExtends.push(config.eslintConfigToExtend);
                     modified = true;
                 }
             });
@@ -207,7 +209,7 @@ export class ConfigureLintingToolsCommand {
             // Create eslintrc
             const eslintrc = {
                 extends: eslintDependencies.map((config) => {
-                    return `${config.eslintrcExtend}`;
+                    return `${config.eslintConfigToExtend}`;
                 })
             };
 
--- a/src/commands/lint/configureLintingToolsCommand.ts
+++ b/src/commands/lint/configureLintingToolsCommand.ts
@@ -11,29 +11,46 @@ import * as path from 'path';
 import { WorkspaceUtils } from '../../utils/workspaceUtils';
 import { JSON_INDENTATION_SPACES } from '../../utils/constants';
 
-const config = workspace.getConfiguration();
-const eslintPluginLwcGraphAnalyzer =
-    '@salesforce/eslint-plugin-lwc-graph-analyzer';
-const eslintPluginLwcGraphAnalyzerConfig =
-    'mobileOfflineLinting.eslint-plugin-lwc-graph-analyzer';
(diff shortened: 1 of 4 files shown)


src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -64,7 +81,7 @@ export class ConfigureLintingToolsCommand {
 
             // Ask user to add eslint plugin
             const result = await this.showMessage(
-                'Do you want to add the ESLint plugin for LWC graph analysis to your project? This will give you linting feedback on code patterns that will not support your LWCs working offline, for mobile use cases.',
+                'Do you want to add the ESLint plugins related to LWC offline capabilities to your project? This will give you linting feedback on code patterns that will not support your LWCs working offline, for mobile use cases.',
```
ben-zhang-at-salesforce, line 84: Hi @khawkins, need your better wording for this msg, thx

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -11,29 +11,46 @@ import * as path from 'path';
...
-    [eslintPluginLwcGraphAnalyzer, eslintPluginLwcGraphAnalyzerVersion],
-    [eslint, eslintVersion]
-];
 
-const lwcGraphAnalyzerRecommended: string =
-    'plugin:@salesforce/lwc-graph-analyzer/recommended';
-const eslintRecommended = 'eslint:recommended';
+const config = workspace.getConfiguration();
+
+class EslintDependencyConfig {
```
khawkins, line 19: In TypeScript, these kinds of data structures are better represented as types (or [type aliases](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-aliases), technically). I actually like their types [cheat sheet](https://www.typescriptlang.org/static/TypeScript%20Types-ae199d69aeecf7d4a2704a528d0fd3f9.png), once your head is around the basic concept. Nice go-to reference for syntax.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -11,29 +11,46 @@ import * as path from 'path';
...
+    readonly name: string;
+    readonly packageJsonPropertyId: string;
+    readonly version: string;
+    readonly eslintrcExtend: string;
```
khawkins, line 23: Property name change suggestion: `eslintConfigToExtend`.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -11,29 +11,46 @@ import * as path from 'path';
...
-];
 
-const lwcGraphAnalyzerRecommended: string =
-    'plugin:@salesforce/lwc-graph-analyzer/recommended';
-const eslintRecommended = 'eslint:recommended';
+const config = workspace.getConfiguration();
+
+class EslintDependencyConfig {
+    readonly name: string;
+    readonly packageJsonPropertyId: string;
```
khawkins, line 21: Property name change suggestion: `packageConfigPropertyId`.

src/commands/lint/configureLintingToolsCommand.ts
```diff
@@ -64,7 +81,7 @@ export class ConfigureLintingToolsCommand {
 
             // Ask user to add eslint plugin
             const result = await this.showMessage(
```
khawkins, line 83: I wonder: do we want to make this all or nothing, now that we're adding multiple plugins? Thinking about it, I guess we can start here, and we can always change it to ask for each plugin, if we get user pushback. KISS to start, works for me.

src/commands/lint/configureLintingToolsCommand.ts
khawkins, line 84:
```suggestion
                'Do you want to add Salesforce code linting guidance for Mobile and Offline capabilities? These tools will identify code patterns that cause problems in Mobile and Offline use cases.',
```

khawkins (commented):
A few suggestions, but overall this looks fine to me.

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: sure, let me do that

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: sure

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: sure

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: or a window with three checkboxes instead of asking three times? should be possible I guess, not sure about effort yet as newbie.

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: sure, nice wording as always.

src/commands/lint/configureLintingToolsCommand.ts
khawkins: Let's do it in one shot like you are here. We can always get fancy if/when we run into scenarios where users want the granularity.

khawkins: (There are a number of different ways, but there's also a fair amount of UX corner case consideration to implement them, so I'd rather not jump all the way into that deep end.)

ben-zhang-at-salesforce: oh, just thought a bit more using checkboxes will that it will more complex due to user could launch the wizard again and then we need to mark the checkbox checked,  ...  for sure more work.  based on no complains from users even wizard is broken for a while, I will suggest to go easy,  just one msg for 3 eslint plugins.

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: read a bit about type/type aliases, I didn't figure out a way to let type to a member as function.  I think class is still better due to we need to have version value to be dynamically set based.   so uses class with a function getVersion() support it.

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: fixed

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: done

src/commands/lint/configureLintingToolsCommand.ts
ben-zhang-at-salesforce: done

src/commands/lint/configureLintingToolsCommand.ts
khawkins: Yeah, it'd be a bit different in the type setup. Instead of having a "lazy load" method to get the version, you'd simply initialize the version with the logic that it contains. Arguably, that's the only reason the `packageConfigPropertyId` exists. So it would instead look something like this:

```typescript
type EslintDependencyConfig = {
    name: string;
    version: string;
    eslintConfigToExtend: string;
};

const eslintDependencies: EslintDependencyConfig[] = [
    {
        name: '@salesforce/eslint-plugin-lwc-mobile',
        eslintConfigToExtend: 'plugin:@salesforce/lwc-mobile/recommended',
        version: config.get('mobileOfflineLinting.eslint-plugin-lwc-mobile')
    },
    // etc.
];
```

But it's not a big deal. It works fine the way you have it too.

khawkins (approved):

ben-zhang-at-salesforce:
[@salesforce/eslint-plugin-lwc-mobile ](https://www.npmjs.com/package/@salesforce/eslint-plugin-lwc-mobile) 1.0.0 is published. so this pr is ready :)

PR #70, opened 2024-07-16
fix: re-authorization is needed even if the offline app is already authorized

haifeng-li-at-salesforce:
The Salesforce logger looks for `transformStream` by using the relative path `../../lib/loggger/transformStream`.  This function belongs to @salesforce-core. The fix is to modify `bundle:extentions` and make the `salesforce-core` external to the generated `extension.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,126 +1,126 @@
 {
-  "name": "salesforcedx-vscode-mobile",
-  "displayName": "%extension.displayName%",
-  "description": "%extension.description%",
-  "version": "0.3.0",
-  "publisher": "salesforce",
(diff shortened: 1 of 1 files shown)


sfdctaka:
Why is there so much shift in lines in the file diff do you know?

haifeng-li-at-salesforce:
> Why is there so much shift in lines in the file diff do you know?

Yes, it is due to the formatting. The only change is line 92

khawkins (approved):

sfdctaka (approved):
LGTM!

PR #114, opened 2024-09-20
W-16323982 - Command for tooling hub Komaci Offline Analysis

angelamuliu:
⚠️ ### **DO NOT MERGE UNTIL CORE RELEASE!**

### What does this PR do?
- Adds command to open a browser tab navigating to the tooling hub tool -> Komaci Offline Analysis -> with arguments for lwc name, namespace piped in
- Command shows up only for LWC relevant files in editor, explorer, and command palette (while file is open)
- Instruments above command by extending CoreExtensionService to pull in telemetry service
- Integration tests

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001xNomIYAS/view


src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,74 @@
...
+            lwcNamespace = 'c'; // Default "c" custom namespace
+        }
+
+        telemetryService.sendCommandEvent(commandName, process.hrtime(), { lwcName, lwcNamespace });
+
+        const workspace = await CoreExtensionService.getWorkspaceContext();
+    
+        // Deploy LWC so analysis runs against current code
+        // TODO: Commented out until TD-0215753 is delivered
+        // const result = await vscode.commands.executeCommand('sf.deploy.current.source.file');
```
sfdctaka, line 40: Curious, where would have `result` be in the following code? It seems as though it's not required to create `toolingHubUrl`?

sfdctaka (approved):
LGTM!

sfdctaka:
@angelamuliu Can you run prettier please.

src/commands/toolingHub/liveKomaciAnalyze.ts
angelamuliu: Result was originally going to be a boolean - true if deploy success, false if deploy false. This is in the air though, maybe they'd want to return a map with a status code.

It isn't necessary to build the url however - because we grab everything we need from the context (which file is open, namespace, etc).

angelamuliu:
@sfdctaka I ran prettier:write and verify, both passed locally. I see the CICD is having issues with it though? Lmk if this is something to be worried about

sfdctaka:
@angelamuliu I made a PR to your branch. On my machine there was one file that needed to be formatted. Maybe that will get over this CI failures? I am certain that formatting will pass, however, this test failure I was also getting on my local machine. Do you see this too?
image

khawkins:
> @angelamuliu I made a PR to your branch. On my machine there was one file that needed to be formatted. Maybe that will get over this CI failures? I am certain that formatting will pass, however, this test failure I was also getting on my local machine. Do you see this too? image

I'm getting the test failure too. Wasn't sure if we just weren't at that point yet or not, but +1.

package.nls.json
```diff
@@ -6,6 +6,7 @@
     "extension.commands.salesforce-mobile-offline.lwc-mobile.version": "Version of ESLint Plugin LWC Mobile to include in devDependencies",
     "extension.commands.salesforce-mobile-offline.komaci.version": "Version of ESLint Plugin LWC Graph Analyzer to include in devDependencies",
     "extension.commands.salesforce-mobile-offline.eslint.version": "Version of ESLint to include in devDependencies",
+    "extension.commands.salesforce-mobile-offline.toolinghub.liveKomaciAnalysis": "(TODO) Salesforce Mobile Offline: Live Komaci Analysis",
```
khawkins, line 9: Nit: This is where you make use of the `category` configuration option in `package.json`, similar to the [linting command](https://github.com/salesforce/salesforcedx-vscode-mobile/blob/e053630226add27c9d75a8c4599306625f95c63e/package.json#L55). So you'd take out the "Salesforce Mobile Offline:" part here, and let the category value/configuration add the prefix.

khawkins, line 9: Nit: go with kebab case for `toolinghub` (`tooling-hub`), for consistency.

package.json
```diff
@@ -53,6 +69,10 @@
                 "command": "salesforcedx-vscode-offline-app.configureLintingTools",
                 "title": "%extension.commands.config-linting-tools.title%",
                 "category": "%extension.commands.salesforce-mobile-offline.category%"
+            },
+            {
+                "command": "salesforcedx-vscode-offline-app.liveKomaciAnalysis",
+                "title": "%extension.commands.salesforce-mobile-offline.toolinghub.liveKomaciAnalysis%"
```
khawkins, line 75: Nit: go with kebab case for `toolinghub` (`tooling-hub`), for consistency.

src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const commandName = 'salesforcedx-vscode-offline-app.liveKomaciAnalysis';
+
+export class LiveKomaciAnalyzeCommand {
+    static async execute(
+        sourceUri: vscode.Uri | vscode.Uri[] | undefined,
+        uris: vscode.Uri[] | undefined
+    ) {
+        const telemetryService =
+            await CoreExtensionService.getTelemetryService();
```
khawkins, line 20: `CoreExtensionService.getTelemetryService()` is a synchronous function. No `await` necessary.

src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        // Extract LWC Name
+        if ((uris && uris?.length > 1) || Array.isArray(sourceUri)) {
```
khawkins, line 23: `(uris && uris?.length > 1)` could be simplified to remove the first condition, and just use:  `uris?.length > 1`.

src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,96 @@
...
+            telemetryService.sendException(
+                commandName + '.invalid_analysis_input',
+                '(TODO) This command runs on a single component.'
+            );
+            return await window.showErrorMessage(
+                '(TODO) This command runs on a single component.'
```
khawkins, line 29: Ideally we wrap all strings shown to the user, in `vscode.l10n.t()`, to provide for localization into other languages in the future.

src/commands/toolingHub/liveKomaciAnalyze.ts
khawkins, line 23: I'm curious: if we don't want multiple values, why are we providing input parameters that allow for multiple values?

package.nls.json
angelamuliu: I had that originally actually - but while developing, I noticed it doesn't append the category consistently? I forget exactly which one out of command palette, editor, and context, but one or two of them was missing category.

I figure this is why the base project opts to have "SFDX:" coded into the command label itself.
https://github.com/forcedotcom/salesforcedx-vscode/blob/232e5af3d5d18d4816338e891175ade9bc3f37bf/packages/salesforcedx-vscode-core/package.nls.json

src/commands/toolingHub/liveKomaciAnalyze.ts
angelamuliu: It's specifically when the command is invoked via context (aka right hand side folder/file browser), and the user selected like...two different LWC html files.

angelamuliu: Unfortunately if I do this, TS ESLINT complains with: 'uris.length' is possibly 'undefined'.ts(18048) due to the comparison following it

there's stuff I can do to work around it, but it's more verbose than above. Hope it's okay to leave as above

khawkins: Ah, interesting. I cannot recreate that diagnostic if I have the following line:

```typescript
if (uris?.length > 1 || Array.isArray(sourceUri)) {
    // ...
```

But I admit that I was misremembering some seemingly nonexistent reality where optional chained "scalar" numeric types evaluated to zero when their parent was undefined. In reality for the case here, while the expression `uris?.length > 1` will evaluate to `false`, it's because `undefined` is not greater than 1. And that's probably not what we want, even if the condition outcome is the same.

TL;DR yes, let's just leave it. 🙂

src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,95 @@
...
+        workspace.getConnection().then((conn) => {
+            let toolingHubUrl =
+                conn.instanceUrl +
+                '/lwr/application/amd/0/e/native/ai/lightningmobileruntime%2Ftoolinghub';
+            toolingHubUrl += `?tab=audit&hideContainer=true&lwcName=${lwcName}&lwcNamespace=${lwcNamespace}&hideForm=true&visualize=true`;
+            vscode.env.openExternal(toolingHubUrl as any);
+        });
+    }
+
+    static getUriFromActiveEditor = (): vscode.Uri | undefined => {
```
khawkins, line 66: Let's be consistent with the format of function definitions, unless there's a specific binding consideration here that I missed. The rest of the functions are defined as methods rather than arrow function properties.

src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,95 @@
...
+        const editor = vscode.window.activeTextEditor;
+        if (editor && editor.document.languageId !== 'forcesourcemanifest') {
```
khawkins, line 68: Am I reading this correctly that we're only looking for the URIs of non-XML LWC files? It might be worth a brief comment here on the intention.

src/commands/toolingHub/liveKomaciAnalyze.ts
```diff
@@ -0,0 +1,95 @@
...
+            return editor.document.uri;
+        }
+        return undefined;
+    };
+
+    static extractLwcNameFromSourceUri(
+        sourceUriInput: vscode.Uri | undefined
+    ): string {
+        const path = sourceUriInput?.path ? sourceUriInput?.path : null;
```
khawkins, line 77: Nit: nullish coalescing would be cleaner here:

```typescript
const path = sourceUriInput?.path ?? null;
```

src/commands/toolingHub/liveKomaciAnalyze.ts
angelamuliu: I'm going to remove this, it's not necessary since I later added file filtering (see package.json where command only shows up for js/ts/css/html)

But for extra context, this code is only triggered when accessing via command palette, and uri for the other files are things like "javascript" and "html." I originally copy pasted this snippet from the deploy command so unsure why they didn't want to include the js-meta file

khawkins:
Going to close this for the time being, as we're going to take a different approach for supporting its service.

PR #127, opened 2024-10-12
Mobile offline

sfdctaka:


package.json
```diff
@@ -84,14 +84,15 @@
     },
     "scripts": {
         "clean": "rimraf out ./lsp/server/out ./lsp/client/out",
-        "compile": "rimraf ./lsp/server/tsconfig.tsbuildinfo ./lsp/client/tsconfig.tsbuildinfo && tsc -b ./",
+        "compile": "rimraf ./lsp/server/tsconfig.tsbuildinfo ./lsp/client/tsconfig.tsbuildinfo && npm run copy-resources && tsc -b ./",
+        "copy-resources": "copyfiles -s -u 3 'lsp/server/resources/*' 'lsp/server/out/resources/'",
```
sfdctaka, line 88: Using `copyfiles` to copy a resource file to the output folder. `tsc` doesn't do this. And we need a utility like `copyfiles` to work on all platforms(Linux/Mac/Windows).

lsp/server/resources/component-experiences.yaml
ben-zhang-at-salesforce, line 1: where is this file from? copy from some repo? how frequent will we refresh this file?

lsp/server/src/validateMobileOffline.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export async function validateMobileOffline(
+    textDocument: TextDocument
+): Promise<Diagnostic[]> {
+    const nonOfflinebaseComponents = getKeysWithoutSpecificValue(
+        baseComponentsAttributes.values,
+        'MobileOffline'
+    );
+    const content = textDocument.getText();
+    const htmlDocument = parseHTMLContent(content);
```
ben-zhang-at-salesforce, line 124: what will happen if parseHTMLContent throw exception when parsing invalid html? our experience on rules for js or graphql is that it crashes the lsp when failed to parse json or graphql

lsp/server/src/validateDocument.ts
```diff
@@ -25,25 +26,26 @@ export async function validateDocument(
...
+
         // handle graphql rules
         const graphqlDiagnostics = await validateGraphql(document);
 
         results = results.concat(jsDiagnostics, graphqlDiagnostics);
-        results.splice(setting.maxNumberOfProblems);
     }
 
     if (document.languageId === 'html') {
+        const mobileOfflineDiagnostics = await validateMobileOffline(document);
```
haifeng-li-at-salesforce, line 43: It would be better if we have a `validateHTML(document: Document')` and a separate `mobile-offline-tag` rule under `diagnostic/html` folder. It will open the door for future diagnostic suppressing. Also it would be easier to maintain in case we will add more rule to html files

lsp/server/src/validateDocument.ts
```diff
@@ -25,25 +26,26 @@ export async function validateDocument(
...
-
-    const setting = await getDocumentSettings(uri);
+    const setting = await getDocumentSettings(document.uri);
     let results: Diagnostic[] = [];
 
     if (document.languageId === 'javascript') {
         // handles JS rules
         const jsDiagnostics = await validateJs(document);
-        
+
```
haifeng-li-at-salesforce, line 35: The result number calculation and truncating need to happen immediately after `validateJS` is called.

package.json
khawkins: You don't really need to do this on all platforms though, do you? This is in the service of developing the extension, not using it. We're not strictly required to support Windows as a development platform for the extension.

lsp/server/src/validateDocument.ts
sfdctaka: Why not cull after all validation?

lsp/server/src/validateDocument.ts
sfdctaka: How much do you want this? I don't mind making the change. But I have to note that the way diagnostics is setup for JS and graphQL is that rules are implementing interface `DiagnosticProducer<T>` which itself declares `validateDocument(...)` . The parameter takes `astNode:T` which is designed to parse AST. Whereas, for html element, the single parameter is `TextDocument`. We can force to make the validators scaffold alike but will soon find out that they're different.

Another argument, and less convincing, I would make is that since this is the one and only html validator still I think future proofing like this isn't strongly required.

That is why I ask, how much do you want this to be done?

package.json
sfdctaka: We need to do this because when lsp is initialized only at that time the yaml file is read in. So the yaml file needs to be packaged.

khawkins: Yes, but it only needs to be packaged at the time of creating the package. We don't have a requirement for a cross-platform way to create the package, just consume it. You could get away with a `cp` here, given that our package creation will almost certainly happen in a *nix environment.

lsp/server/src/validateDocument.ts
haifeng-li-at-salesforce: If JS validator exceeds maximum, we don't need to do the GQL validator

lsp/server/src/validateDocument.ts
haifeng-li-at-salesforce: param `astNode:T` within `validateDocument` is confusing. It should be able to handle any date structure since it is generic. `TextDocument` should be fine.  It is better to put `mobile-offline` as ` rule since we take take the advantage of `DiagnosticProducer` to support suppressing

lsp/server/src/validateDocument.ts
sfdctaka: Maximum threshold to me is a way to cull the results array before showing them to the user. And not for early exit. If what you're saying is what we want to achieve, then, it is implied that we cull every time we call `validateXYZ(...)`. If that's the preferred approach I can change.

lsp/server/src/validateDocument.ts
sfdctaka: That's what I'm saying. Validating for mobile offline DOESN'T USE DiagnosticProducer. I'm just asking to make sure we're not mashing up things for the sake of everything need to follow the same approach.

haifeng-li-at-salesforce: What I am suggesting is to use `DiagnosticProducer` for html files to keep the pattern the same. We can change the signature of `DiagnosticProducer` interface to export interface DiagnosticProducer<T> {
    ```
/**
     * Validate the parsed text document as astNode and return a list of diagnostics.
     * @param textDocument the language server text document
     * @param data  data parsed from document body
     * @returns An array of diagnostics found within document
     */
    validateDocument(
        textDocument: TextDocument,
        data: T
    ): Promise<Diagnostic[]>;
}

```

Comment (sfdctaka) [reply lsp/server/src/validateDocument.ts:43]: But mobile offline validator doesn't need `data:T`. That's someone else's concern. You're trying to mashup method signature when there is no need for the sake of trying to look alike.

Comment (haifeng-li-at-salesforce) [reply lsp/server/src/validateDocument.ts:43]: The need is for future suppressing support and each `DiagnosticProducer` is going to have an ID for fine tuning validator suppression. For example, in `setting.json`, we can define `“mobile.lsp.diagnostic”: {
            “suppress.all”: “true”,
            “suppressed.rule.Ids”: [“adapters-local-change-not-aware”, "mobile-offline"],
            “maxProblemNumber”: 500
 }`

If you don't like to create another `DiagnosticProducer`, we can come up with other approach to support it

Comment (ben-zhang-at-salesforce) [reply lsp/server/src/validateDocument.ts:43]: it will be great to has validateHtml.ts and move the offline individual into diagnostic rule folder.  this moment it's end of world.  I can do the re-arrangement when I will merge your change into the suppressing piece I am doing. 

so for this one, take as is?

Comment (sfdctaka) [reply lsp/server/src/validateDocument.ts:43]: @haifeng-li-at-salesforce where is the rule ID defined? I don't see it. If there's a valid reason as you stated, like I said, I'm not against changing. But with the way your code has been written I don't know how `"suppressed.rule.Ids"` are supposed to work. Is this something you were planning on doing it later?

Comment (ben-zhang-at-salesforce) [reply lsp/server/src/validateDocument.ts:43]: yes, I am working on the suppressing stuff now, hope to have the pr tomorrow morning.  we can huddle and show what the implementation.

Comment (sfdctaka) [reply lsp/server/src/validateDocument.ts:43]: I would never be able to know what you and/or Haifeng are preparing now unless it is already in the codebase. And to be asked for something that's not there is harder. Let's huddle tomorrow. If I'm understanding you correctly there's a work you're working on now that should come before this(my) PR then.

Comment (ben-zhang-at-salesforce) [reply lsp/server/src/validateDocument.ts:43]: Hi @sfdctaka I am sorry I commented before showing the diagnostic pr.  now the [preview pr](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/135) is up.

File: lsp/server/src/diagnostic/html/mobileOfflineFriendly.ts [hunk 5]
```diff
@@ -0,0 +1,162 @@
...
+export class MobileOfflineFriendly implements DiagnosticProducer<HTMLDocument> {
+    getId(): string {
+        return RULE_ID;
+    }
+
+    async validateDocument(
+        textDocument: TextDocument,
+        data: HTMLDocument
+    ): Promise<Diagnostic[]> {
+        const nonOfflinebaseComponents = getKeysWithoutSpecificValue(
```
ben-zhang-at-salesforce, line 32:
```suggestion
        const nonOfflineBaseComponents = ...
```

lsp/server/src/diagnostic/html/mobileOfflineFriendly.ts
```diff
@@ -0,0 +1,162 @@
...
+        str
+            // Replace uppercase letters with a dash followed by the lowercase letter
+            .replace(/([A-Z])/g, '-$1')
+            // Convert the entire string to lowercase
+            .toLowerCase()
+    );
+}
+
+/**
+ * Parse componenent-experiences.yaml come in a format like this
```
ben-zhang-at-salesforce, line 92: nitpick: 
```suggestion
 * Parse component-experiences.yaml come in a format like this
```

lsp/server/src/validateDocument.ts
```diff
@@ -26,26 +26,25 @@ export async function validateDocument(
...
-        );
+        const graphqlDiagnostics = await validateGraphql(setting, document);
 
         results = results.concat(jsDiagnostics, graphqlDiagnostics);
     }
 
     if (document.languageId === 'html') {
+        const diagnostics = await validateHtml(setting, document);
+        results = results.concat(diagnostics);
+        var aaa = 0;
```
ben-zhang-at-salesforce, line 46: nitpick: could be removed

lsp/server/src/validateHtml.ts
ben-zhang-at-salesforce, line 1: copyright is missing here

lsp/server/src/validateHtml.ts
```diff
@@ -0,0 +1,48 @@
...
+    textDocument: TextDocument
+): Promise<Diagnostic[]> {
+    let results: Diagnostic[] = [];
+
+    const producers = diagnosticProducers.filter((producer) => {
+        return !isTheDiagnosticSuppressed(setting, producer.getId());
+    });
+
+    if (producers.length > 0) {
+        try {
```
ben-zhang-at-salesforce, line 31: Today such try catch block too big issue is noticed.  It swallows the exception from the parseHtml call exception and calls to await producer.validateDocument.   If there are more than one producers, and if the first producer throws, the second producer will be not be called ... 
We have similar issue in validateJs and valideGraphql.  Time to abstract and centralize?

lsp/server/package.json
```diff
@@ -23,6 +24,10 @@
         "@babel/preset-env": "^7.25.7",
         "@types/babel__traverse": "7.20.6",
         "@types/babel-core": "6.25.10",
-        "graphql": "^16.9.0"
+        "@types/mocha": "^10.0.7",
+        "@types/sinon": "^17.0.3",
```
haifeng-li-at-salesforce, line 28: child project could reference root project's modules. So `@type/mocha` and `@types/sinon` is not necessary here.

lsp/server/src/diagnostic/html/mobileOfflineFriendly.ts
```diff
@@ -0,0 +1,162 @@
...
+// import traverse from '@babel/traverse';
+import { Node, HTMLDocument } from 'vscode-html-languageservice';
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import { DiagnosticProducer } from '../DiagnosticProducer';
+import { Diagnostic, DiagnosticSeverity } from 'vscode-languageserver/node';
+import { transformYamlToObject } from '../../utils/yamlParser';
+import * as path from 'path';
+import * as fs from 'fs';
+
+// const SEVERITY = DiagnosticSeverity.Information;
```
haifeng-li-at-salesforce, line 18: remove

lsp/server/src/diagnostic/html/mobileOfflineFriendly.ts
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+// import { Node, isCallExpression } from '@babel/types';
+// import traverse from '@babel/traverse';
```
haifeng-li-at-salesforce, line 9: remove

lsp/server/src/validateDocument.ts
sfdctaka: Argh. You are right, it should be removed. Left over from debugging the code.

lsp/server/src/validateHtml.ts
sfdctaka: I followed your code pattern. And this is one of the reasons why we shouldn't prescribe to unproven pattern or scaffolding because it's likely that these antipatterns emerge.

Your suggestion is valid though. But it shouldn't be done for this PR.

haifeng-li-at-salesforce (approved):
Code looks great!

ben-zhang-at-salesforce (approved):
look very good, great job!

PR #129, opened 2024-10-15
feat: use workspace dependency 

haifeng-li-at-salesforce:
### What does this PR do?

Treat lsp/client and lsp/server as sub-project in workspace.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,9 @@
     "description": "%extension.description%",
     "version": "0.3.0",
     "publisher": "salesforce",
+    "workspaces": [
+        "lsp/*"
+    ],
     "engines": {
         "vscode": "^1.77.0",
         "node": ">=18"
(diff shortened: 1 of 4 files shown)


sfdctaka (approved):
LGTM!

package.json
```diff
@@ -123,6 +125,8 @@
         "@salesforce/core": "^5.3.12",
         "@salesforce/lwc-dev-mobile-core": "^3.3.1",
         "@salesforce/sf-plugins-core": "^4.0.0",
-        "@vscode/l10n": "^0.0.18"
+        "@vscode/l10n": "^0.0.18",
+        "mobile-lsp-client": "workspaces:*",
+        "mobile-lsp-server": "workspaces:*"
```
maliroteh-sf, line 130: @haifeng-li-at-salesforce @sfdctaka 

Since we have `lsp/client` and `lsp/server` folders, wouldn't it be better is we define 2 workspaces (one for client and one for server) instead of `"workspaces": [ "lsp/*" ]` then in here we would individually reference those workspaces?

As it is right now, both `mobile-lsp-client` and `mobile-lsp-server` would mean the same thing, right? So `import * as lspClient from 'mobile-lsp-client';` and `import * as lspClient from 'mobile-lsp-servert';` would be the same, which doesn't make sense.

haifeng-li-at-salesforce: 'mobile-lsp-client' refers to the package name defined within 'lsp/client/package.json'. So it case for 'mobile-lsp-server'. 'workspace:*' means that package dependency would look for dependency 'mobile-lsp-client' under 'workspace', in our case 'lsp'.
So `import * as lspClient from 'mobile-lsp-client'; and import * as lspClient from 'mobile-lsp-servert'` are different.

haifeng-li-at-salesforce: But I agree that using 'workspace:server' or 'workspace:client' is more intuitive. I will include the suggestion in next PR.

PR #136, opened 2024-10-18
Utilize workspace settings and code action to let developer configure or suppress diagnostics

ben-zhang-at-salesforce:
We will implement lsp diagnostics, like component offline support checking, graphql or js mobile related ... Developer needs the capability to suppress them.

This pr does:

two Code Action quickfix: suppress one type of diagnostics or suppress all lds mobile diagnostic.
limit the max diagnostics to return


lsp/server/src/diagnostic/DiagnosticSettings.ts
```diff
@@ -0,0 +1,62 @@
+    /*
```
sfdctaka, line 1: Let's run prettier.

lsp/server/src/diagnostic/DiagnosticSettings.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Take in input, possibly a diagnostic json setting true and get valid settings 
+ * @param input the input, a well format tree is like 
+ * 
+ *  {
+            "salesforce.salesforcedx-vscode-mobile.diagnostics.suppressAll": true,
+            "salesforce.salesforcedx-vscode-mobile.diagnostics.suppressedIds": ["adapters-local-change-not-aware"],
+            "salesforce.salesforcedx-vscode-mobile.diagnostics.maxProblemNumber": 50
+    }
+ * 
+ * @returns
```
sfdctaka, line 47: Missing returns description

lsp/server/src/diagnostic/gql/misspelled-uiapi.ts
```diff
@@ -19,6 +20,10 @@ const SEVERITY = DiagnosticSeverity.Error;
 */
 export class MisspelledUiapi implements DiagnosticProducer<ASTNode> {
    
+    getId(): DiagnosticId {
+        return 'missspelled-uiapi' ;
```
sfdctaka, line 24: Missspelled is misspelled :)

lsp/server/src/settings.ts
```diff
@@ -0,0 +1,36 @@
+
```
sfdctaka, line 1: Copyright

lsp/server/src/server.ts
```diff
@@ -30,10 +34,21 @@ let hasConfigurationCapability = false;
 let hasWorkspaceFolderCapability = false;
 export let hasDiagnosticRelatedInformationCapability = false;
 
+let extensionId: string = '';
```
sfdctaka, line 37: Can we get rid of explicit typing like `: string` when we are initializing it to empty string? In TypeScript, implicit type declaration (also known as type inference) occurs when you declare a variable and assign it a value without explicitly specifying the type. TypeScript automatically infers the type based on the assigned value.

lsp/server/src/settings.ts
```diff
@@ -0,0 +1,36 @@
...
+// Please note that this is not the case when using this server with the client provided
+// but could happen with other clients.
+export const defaultSettings: Settings = { 
+    diagnostic: defaultDiagnosticSettings
+};
+
+export let globalSettings: Settings = defaultSettings;
+
+/**
+ *
```
sfdctaka, line 24: Missing jsdoc content.

lsp/server/src/diagnostic/DiagnosticProducer.ts
```diff
@@ -7,8 +7,15 @@
...
 import { Diagnostic } from 'vscode-languageserver/node';
 import { TextDocument } from 'vscode-languageserver-textdocument';
+import { DiagnosticId } from './DiagnosticSettings';
 
 export interface DiagnosticProducer<T> {
+
+    /**
+     * Get the Id for the diagnostic producer. 
+     */
+    getId(): DiagnosticId;
```
ben-zhang-at-salesforce, line 17: each individual producer will need to return a Id

lsp/server/src/diagnostic/DiagnosticProducer.ts
```diff
@@ -19,4 +26,9 @@ export interface DiagnosticProducer<T> {
         textDocument: TextDocument,
         astNode: T
     ): Promise<Diagnostic[]>;
+
+}
+
+export interface DiagnosticMetaData {
```
ben-zhang-at-salesforce, line 32: Diagnostic sent from server to client could have data?: Any.  we utilize it to send the producer id, so CodeAction could implement a quickfix to suppress that producer.

lsp/server/src/server.ts
```diff
@@ -75,7 +91,7 @@ connection.onInitialized(() => {
         // Register for all configuration changes.
         connection.client.register(
             DidChangeConfigurationNotification.type,
-            undefined
+            { section: diagnosticsSettingSection }
```
ben-zhang-at-salesforce, line 94: here we register to watch the change happens only under 'salesforce.salesforcedx-vscode-mobile.diagnostics' section.

lsp/server/src/server.ts
```diff
@@ -85,65 +101,40 @@ connection.onInitialized(() => {
...
-    maxNumberOfProblems: number; //max number of diagnostics to detect per document.
-}
-
-// The global settings, used when the `workspace/configuration` request is not supported by the client.
-// Please note that this is not the case when using this server with the client provided
-// but could happen with other clients.
-const defaultSettings: MobileSettings = { maxNumberOfProblems: 1000 };
-let globalSettings: MobileSettings = defaultSettings;
+connection.onDidChangeConfiguration((change) => {
+    const changedSetting =
```
ben-zhang-at-salesforce, line 105: when settings under salesforce.salesforcedx-vscode-mobile.diagnostics are changed, get the new settings.

lsp/server/src/server.ts
```diff
@@ -155,6 +146,58 @@ connection.languages.diagnostics.on(async (params) => {
...
+	}
+    
+	const result: CodeAction[] = [];
+    diagnostics.forEach((item) => {
+        const { data } = item;
+        if (data !== undefined) {
+            const metData = data as DiagnosticMetaData;
+            if (metData.id !== undefined) {
+
+                const suppressedIds = new Set(settings.diagnostic.suppressedIds);
```
ben-zhang-at-salesforce, line 163: generate the quickfix to suppress this diagnostic producer.

lsp/server/src/server.ts
```diff
@@ -155,6 +146,58 @@ connection.languages.diagnostics.on(async (params) => {
...
+                        title: 'Update workspace setting',
+                        command: updateDiagnosticsSettingCommand,
+                        arguments: [{
+                            suppressedIds: Array.from(suppressedIds)
+                        }]
+                    }
+                };
+                result.push(suppressThisDiagnostic);
+
+                const suppressAllDiagnostic: CodeAction = {
```
ben-zhang-at-salesforce, line 179: quick fix to suppress all diagnostics from vscode-mobile

package.json
```diff
@@ -75,6 +75,21 @@
                     "type": "string",
                     "default": "^8.47.0",
                     "description": "%extension.commands.salesforce-mobile-offline.eslint.version%"
+                },
+                "salesforce.salesforcedx-vscode-mobile.diagnostics.suppressAll": {
```
ben-zhang-at-salesforce, line 79: add 3 diagnostic setting, so developer could manually change it from vscode setting ui.

src/commands/settings/settings.ts
ben-zhang-at-salesforce, line 1: The setting command lsp sever code action will calls into update workspace diagnostic settings. 
This command is headless, could be called programatically, but is not visible when cmd+shift+p is clicked.

src/commands/settings/settings.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export function registerCommand(context: vscode.ExtensionContext) {
+    const command = getUpdateDiagnosticsSettingCommand(context); 
+    context.subscriptions.push(vscode.commands.registerCommand(
+        command,
+        async (diagnosticSetting) => {
+            const section = getDiagnosticsSettingSection(context);
+            const config = vscode.workspace.getConfiguration(section);
+            try {
```
ben-zhang-at-salesforce, line 27: pull the individual setting from the object, if valid, update to workspace

src/extension.ts
```diff
@@ -37,9 +40,20 @@ export function activate(context: vscode.ExtensionContext) {
...
+    settingsCommand.registerCommand(context);
 
     configureLintingToolsCommand.registerCommand(context);
-    lspClient.activate(context);
+
+    const extensionId = getExtensionId(context);
+    const command = getUpdateDiagnosticsSettingCommand(context);
+    const section = getDiagnosticsSettingSection(context);
+
+    lspClient.activate(
```
ben-zhang-at-salesforce, line 51: pass in three 3 strings into client, then server to avoid hardcoding in multiple place.

lsp/server/src/server.ts
```diff
@@ -155,6 +141,58 @@ connection.languages.diagnostics.on(async (params) => {
...
+    diagnostics.forEach((item) => {
+        const { data } = item;
+        if (data !== undefined) {
+            const metData = data as DiagnosticMetaData;
+            if (metData.producerId !== undefined) {
+
+                const suppressedIds = new Set(settings.diagnostic.suppressedIds);
+                suppressedIds.add(metData.producerId);
+                const suppressThisDiagnostic: CodeAction = {
+                    title: `Suppress such diagnostic: ${metData.producerId}`,
```
ben-zhang-at-salesforce, line 161: the action action title is visible to developer like below, need a official wording for it.
image

package.json
ben-zhang-at-salesforce: the are visible to developer like below, will need an official wording for them:  the id and description.
image

lsp/server/src/diagnostic/DiagnosticProducer.ts
```diff
@@ -19,4 +26,9 @@ export interface DiagnosticProducer<T> {
         textDocument: TextDocument,
         astNode: T
```
sfdctaka, line 27: Like we discussed in the other [PR](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/127#discussion_r1805518966) you should make `astNode` a more generic name and probably a optional parameter.

lsp/server/src/diagnostic/DiagnosticSettings.ts
```diff
@@ -0,0 +1,62 @@
+    /*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export type DiagnosticId = 
+    | 'missspelled-uiapi' 
+    | 'adapters-local-change-not-aware';
```
sfdctaka: I understand why you have this `type`. But diagnostic producers in their respective `getId()` also returns hard coded string. Ideally, self-declaring ID should be enough. 

With the way the code is written now, as we see it here, we need to register the string because we want to create something like `Set<DiagnosticId>` in line 18 below. And for this reason I prefer something more like
```
export type DiagnosticId = string;
```
I think you can do this in Typescript but need to verify.

lsp/server/src/server.ts
```diff
@@ -155,6 +141,58 @@ connection.languages.diagnostics.on(async (params) => {
...
 });
 
+connection.onCodeAction((params) => {
+    const textDocument = documentCache.get(params.textDocument.uri);
+    const diagnostics = params.context.diagnostics;
+	if (textDocument === undefined || diagnostics.length === 0) {
+		return undefined;
+	}
+    
+	const result: CodeAction[] = [];
```
sfdctaka, line 151: Format

lsp/server/src/server.ts
```diff
@@ -155,6 +141,58 @@ connection.languages.diagnostics.on(async (params) => {
...
+    diagnostics.forEach((item) => {
+        const { data } = item;
+        if (data !== undefined) {
```
sfdctaka, line 154: nit:  this reads much easier
```
diagnostics.forEach((diagnostic) => {
    if (diagnostic.data !== undefined) {
```

lsp/server/src/server.ts
sfdctaka: A `Set` created from suppressed Ids don't need to be created in each loop of `forEach` does it?

lsp/server/src/diagnostic/DiagnosticSettings.ts
ben-zhang-at-salesforce: done

lsp/server/src/diagnostic/gql/misspelled-uiapi.ts
ben-zhang-at-salesforce: corrected, good eye

lsp/server/src/server.ts
ben-zhang-at-salesforce: done

lsp/server/src/diagnostic/DiagnosticProducer.ts
ben-zhang-at-salesforce: sure, I will name it 'data'

lsp/server/src/diagnostic/DiagnosticSettings.ts
ben-zhang-at-salesforce: Yes, I can use export` type DiagnosticId = string;`.  personally I 
like ```type DiagnosticId = 
    | 'missspelled-uiapi' 
    | 'adapters-local-change-not-aware'; 
```. it prevents a ramdam string is returned if forget to reference from the central place.  This pattern is used extensively in lds.  

if you still think **export type DiagnosticId = string;** is better, I can change it.

File: package.json (hunk 13 above)
Comment (khawkins) [inline package.json:79]: These names should be trimmed, as they translate directly into titles and other descriptive verbiage in the settings UI. Taking some cues from `properties` section of [Configuration schema](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema), it feels like a better name for this property would be:

```suggestion
                "salesforcedx-vscode-mobile.diagnostics.suppressAll": {
```

Which I _think_ should translated in the UI to:

# Salesforce Mobile Extensions

_Other settings for the extension_

**Diagnostics: Suppress All**
_Description, input, etc._

package.json
```diff
@@ -75,6 +75,21 @@
...
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Suppress all problem detection."
```
khawkins, line 82:
```suggestion
                    "description": "Suppress all problem detection"
```

package.json
```diff
@@ -75,6 +75,21 @@
...
+                },
+                "salesforce.salesforcedx-vscode-mobile.diagnostics.suppressedIds": {
```
khawkins, line 84:
```suggestion
                "salesforcedx-vscode-mobile.diagnostics.supressByRuleId": {
```

package.json
```diff
@@ -75,6 +75,21 @@
...
+                    "type": "array",
+                    "default": [],
+                    "description": "Suppress problem by Id(s)."
```
khawkins, line 87:
```suggestion
                    "description": "Suppress problem detection specifically for one or more rule IDs"
```

package.json
```diff
@@ -75,6 +75,21 @@
...
+                },
+                "salesforce.salesforcedx-vscode-mobile.diagnostics.maxProblemNumber": {
```
khawkins, line 89:
```suggestion
                "salesforcedx-vscode-mobile.diagnostics.maxNumberOfProblems": {
```

package.json
```diff
@@ -75,6 +75,21 @@
...
+                    "type": "integer",
+                    "default": 200,
+                    "description": "The max number of problem to detect."
```
khawkins, line 92:
```suggestion
                    "description": "The maximum number of detected problems to report"
```

lsp/server/src/diagnostic/DiagnosticSettings.ts
sfdctaka: I like to learn why you like your original approach.

lsp/server/src/validateGraphql.ts
```diff
@@ -53,17 +68,31 @@ export async function validateGraphql(
...
-            )
+            producers.map((producer) => {
+                return producer.validateDocument(textDocument, graphqlAstNode)
+                 .then((diagnostics) => {
+                     const metaData: DiagnosticMetaData = {
+                         producerId: producer.getId()
+                     };
+                     diagnostics.forEach((diagnostic) => {
+                         diagnostic.data = metaData;
+                     });
```
sfdctaka, line 92: I understand why you made `DiagnosticMetaData`, that is, to take advantage of `Diagnostic` object's `data?` property, which unforunately is of type `any`.

But having a metadata that only has one property feels like an overkill. And the post processing of attaching such data on each of the diagnostic makes this argument stronger.

Can we devise of something else?
We can use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). Extend `Diagnostic` interface in the `src/@types` folder
image
Then you'll be able to add a property to `Diagnostic` and get rid of `DiagnosticMetaData`.

ben-zhang-at-salesforce: let me give [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) a try, it will be great if property will be kept while the diagnostic is sent back and force between lsp client and server.

lsp/server/src/settings.ts
ben-zhang-at-salesforce: done

lsp/server/src/diagnostic/DiagnosticSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export type ProducerId = 
+    | 'misspelled-uiapi' 
+    | 'adapters-local-change-not-aware'; 
+
```
haifeng-li-at-salesforce, line 11: Should `ProducerId` be generated at runtime by iterating all the producers belonging to each validator?

lsp/server/src/diagnostic/DiagnosticProducer.ts
```diff
@@ -7,16 +7,28 @@
...
 import { Diagnostic } from 'vscode-languageserver/node';
 import { TextDocument } from 'vscode-languageserver-textdocument';
+import { ProducerId } from './DiagnosticSettings';
 
 export interface DiagnosticProducer<T> {
+
+    /**
+     * Get the Id for the diagnostic producer. 
+     */
+    getId(): ProducerId;
```
haifeng-li-at-salesforce, line 17: Is string type enough?

lsp/server/src/diagnostic/DiagnosticSettings.ts
ben-zhang-at-salesforce: sure! 
image
I like the compiling time error check :)

lsp/server/src/diagnostic/DiagnosticSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+ * 
+ *  {
+            "suppressAll": true,
+            "suppressedIds": ["adapters-local-change-not-aware"],
+            "maxProblemNumber": 50
+    }
+ * 
+ * @returns the settings for diagnostics
+ */
+export function getSettings(currentSetting: DiagnosticSettings, input: any): DiagnosticSettings {
```
haifeng-li-at-salesforce, line 50: Is it better to have only 1 param `input`? All missing parts can be picked up from `defaultDiagnosticSettings`

lsp/server/src/diagnostic/DiagnosticSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (input === undefined) {
+        return currentSetting;
+    }
+    // pull the values from input
+    const maxNumberOfProblems = parseInt(input['maxProblemNumber']);
+    const suppressAll = input['suppressAll'] === true?true:(input['suppressAll'] === false?false:undefined);
```
haifeng-li-at-salesforce, line 56:
```suggestion
    const suppressAll = typeof input['suppressAll'] === 'boolean' ? input['suppressAll'] : undefined;
```

lsp/server/src/server.ts
haifeng-li-at-salesforce: This is smart!

lsp/server/src/diagnostic/DiagnosticSettings.ts
sfdctaka: I don't think I'm fully getting it.

You get misspelling because you defined it in `ProducerId`. 

If you don't define any thing in `ProducerId`, then, producer ID of the diagnostic rule can be anything and be self declaring.

So I don't see the benefit in the argument that you're making. Except that, you liked to be checked by the guard rail that you self imposed. 

If I missed seeing the benefit I'm all ears!

lsp/server/src/diagnostic/DiagnosticProducer.ts
ben-zhang-at-salesforce: @haifeng-li-at-salesforce @sfdctaka I over thought about the enforcing producer id uniqueness by defining the in cetral place. let me remove the ProducerId thing.  hope the individual producer will pick the same id.

sfdctaka: This point that you make, is what I wanted to hear in my other question about the benefit of having union type of string in `ProducerId`. So you want a unique ID for producer.

I still think union type is not the way to go.

I think @haifeng-li-at-salesforce has a great point. Produce ID probably doesn't suffice what's trying to be solved here. The problem is to be able to filter/suppress diagnostics. And to effectively do that in runtime without name collision is to have a unique compile time ID and to use that ID instead of names(string). Simple thinking me says give a static GUID to the diagnostic rule. You can make this a `ProducerId` if you like. And you'll add another property in `Diagnostics` that's used only for displaying name of that diagnostic.

lsp/server/src/validateGraphql.ts
ben-zhang-at-salesforce: tried, unfortunately the produceId is lost :( 
image

I will use data to hold the produceId directly without the Metadata extra structure.  I over thought for future proof it.

lsp/server/src/server.ts
ben-zhang-at-salesforce: done

lsp/server/src/diagnostic/DiagnosticProducer.ts
ben-zhang-at-salesforce: done

package.json
ben-zhang-at-salesforce: done

package.json
ben-zhang-at-salesforce: done

package.json
ben-zhang-at-salesforce: done

package.json
ben-zhang-at-salesforce: done

lsp/server/src/diagnostic/DiagnosticSettings.ts
ben-zhang-at-salesforce: removed.

lsp/server/src/diagnostic/DiagnosticProducer.ts
ben-zhang-at-salesforce: removed.

lsp/server/src/server.ts
```diff
@@ -85,65 +98,37 @@ connection.onInitialized(() => {
...
-// The global settings, used when the `workspace/configuration` request is not supported by the client.
-// Please note that this is not the case when using this server with the client provided
-// but could happen with other clients.
-const defaultSettings: MobileSettings = { maxNumberOfProblems: 1000 };
-let globalSettings: MobileSettings = defaultSettings;
+connection.onDidChangeConfiguration((change) => {
+    const changedSetting =
+        diagnosticsSettingSection
+            .split('.')
+            .reduce((parent, key) => parent[key], change.settings);
```
haifeng-li-at-salesforce, line 106: Can we give an example about how this data transformation works.

lsp/server/src/validateGraphql.ts
sfdctaka: How funny. It worked on my machine.

lsp/server/src/server.ts
```diff
@@ -155,6 +140,54 @@ connection.languages.diagnostics.on(async (params) => {
...
+	if (textDocument === undefined || diagnostics.length === 0) {
+	    return undefined;
+	}
+    
+	const result: CodeAction[] = [];
+
+    diagnostics.forEach((diagnostic) => {
+        // generate the two suppressing quick fixes 
+        const { data : producerId } = diagnostic;
+        if (typeof producerId === 'string') {
```
sfdctaka, line 155: Do you need to check the type of `producerId`?

lsp/server/src/server.ts
```diff
@@ -155,6 +140,54 @@ connection.languages.diagnostics.on(async (params) => {
     }
 });
 
+connection.onCodeAction((params) => {
+    const textDocument = documentCache.get(params.textDocument.uri);
+    const diagnostics = params.context.diagnostics;
+	if (textDocument === undefined || diagnostics.length === 0) {
```
sfdctaka, line 146: Indentation

lsp/server/src/settings.ts
```diff
@@ -0,0 +1,39 @@
...
+ * if the property from input is missing or invalid. 
+ * @param currentSetting current setting
+ * @param input the input to pull setting value from.
+ */
+export function getSettings(currentSetting: Settings, input: any): Settings {
+    const diagnosticSetting = getDiagnosticSettings(currentSetting.diagnostic, input);
+    return {
+        diagnostic: diagnosticSetting
+    }
+}
```
haifeng-li-at-salesforce, line 39: Could we remove this layer of management and  put `globalSettings` and `getSetting` into `DiagnosticSettings`

lsp/server/src/validateGraphql.ts
sfdctaka: If this approach is also another overkill then I'd rather have your updated approach in simple string property in `DiagnosticProducer` 👍

lsp/server/src/server.ts
ben-zhang-at-salesforce: when instantiate a new set, a array like structure could be passed in as constructor parameter.  the items from input parameter will become items of the set.  so no need to loop :)

package.json
ben-zhang-at-salesforce: the max number of problems are removed, team deem it's not required.

src/utils/workspaceUtils.ts
```diff
@@ -136,3 +136,7 @@ export class NoStaticResourcesDirError extends Error {
         Object.setPrototypeOf(this, NoStaticResourcesDirError.prototype);
     }
 }
+
+export function getExtensionName(context: ExtensionContext): string {
```
haifeng-li-at-salesforce, line 140: Could get extension be a string separated by space? Will that cause problem?

lsp/server/src/diagnostic/DiagnosticSettings.ts
ben-zhang-at-salesforce: sure, I over thought about the setting system, let me simplify it.

lsp/server/src/server.ts
```diff
@@ -155,6 +140,54 @@ connection.languages.diagnostics.on(async (params) => {
...
+	    return undefined;
+	}
+    
+	const result: CodeAction[] = [];
+
+    diagnostics.forEach((diagnostic) => {
+        // generate the two suppressing quick fixes 
+        const { data : producerId } = diagnostic;
```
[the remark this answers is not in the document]
ben-zhang-at-salesforce: this line is to read the data property out for diagnostic and name it as a producerId const.

lsp/server/src/server.ts
ben-zhang-at-salesforce: you are right, we don't need it for now, let me remove it

lsp/server/src/server.ts
ben-zhang-at-salesforce: done

lsp/server/src/server.ts
ben-zhang-at-salesforce: done

lsp/server/src/settings.ts
ben-zhang-at-salesforce: done

lsp/server/src/diagnostic/DiagnosticSettings.ts
ben-zhang-at-salesforce: done

lsp/server/src/server.ts
ben-zhang-at-salesforce: added comments explaining how it works.

sfdctaka: Where is the comment?

src/utils/workspaceUtils.ts
ben-zhang-at-salesforce: this is the package.json name attribue, currently it's 'salesforcedx-vscode-mobile'. not like to change it to have space.

lsp/server/src/server.ts
ben-zhang-at-salesforce: I guess I clicked the "Add single comment" before pushing the change, my bad.  pushed now.

lsp/server/src/test/validateDocument.spec.ts
ben-zhang-at-salesforce, line 1: validateDocument.ts is quite simple. the e2e tests from lsp/client/src/test/*.test.ts will cover it better.

lsp/server/src/test/validateGraphql.spec.ts
ben-zhang-at-salesforce, line 1: left over, validateGraphql.test.ts is the correct test file.

lsp/server/src/server.ts
```diff
@@ -155,6 +155,56 @@ connection.languages.diagnostics.on(async (params) => {
...
+    const diagnostics = params.context.diagnostics;
+    if (textDocument === undefined || diagnostics.length === 0) {
+        return undefined;
+    }
+
+    const result: CodeAction[] = [];
+
+    diagnostics.forEach((diagnostic) => {
+        // generate the two suppressing quick fixes
+        const { data: producerId } = diagnostic;
```
sfdctaka, line 169: I still don't understand what this does. Why not use `diagnostic.producerId`?

ben-zhang-at-salesforce: changed  `const { data: producerId } = diagnostic; ` to `const producerId = diagnostic.data;` 

About your question **Why not use diagnostic.producerId?**,  I tried, the producerId doesn't persist cross when it's being shuffled between server to client and client back to server.  Here's the exact steps I did:

1. extended Diagnostic interface to have new property 'produceId' using [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)
2. set diagnostic.producerId when the diagnostic is generated for lsp client to show it. 
3. when mousing over the rendered diagnostic, Server.onCodeAction is called with a parameter of CodeActionParams.  The parameter holds an instance of the diagnostic. the producerId of that diagnostic is undefined.  

Seems that lsp system doesn't like the producerId and clean it away. 

That's my findings.

sfdctaka (approved):
LGTM

package.json
ben-zhang-at-salesforce: thanks @khawkins giving the wording for it.  now it's below. 
image

PR #141, opened 2024-10-21
get object Info for authorized org for LSP

haifeng-li-at-salesforce:
### What does this PR do?
If org is authorized, fetch the object info at runtime and cache it. This is the pre-requisite for LSP rule which requires runtime objectinfo query, like oversized field and oversized records in a gql query.

### What issues does this PR fix or reference?
No


lsp/server/package.json
```diff
@@ -6,6 +6,7 @@
     "engines": {
         "node": "*"
     },
+    "main": "out/index.js",
     "publisher": "salesforce",
```
haifeng-li-at-salesforce, line 10: export dependencies for root project

lsp/server/src/index.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export { TempProjectDirManager } from './test/TestHelper';
```
haifeng-li-at-salesforce, line 8: It used to be in root project, moved here for broader usage

lsp/server/src/test/TestHelper.ts
```diff
@@ -0,0 +1,153 @@
...
+                const projectDir = await mkdtemp(
+                    path.join(os.tmpdir(), 'offlineWizard-')
+                );
+                return resolve(new TempProjectDirManager(projectDir));
+            } catch (err) {
+                return reject(err);
+            }
+        });
+    }
+}
```
haifeng-li-at-salesforce, line 71: `TempProjectDirManager` used to be in root project

lsp/server/src/types/FieldRepresentation.ts
```diff
@@ -0,0 +1,35 @@
...
+    referenceTargetField: string | null;
+    referenceToInfos: any;
+    relationshipName: string | null;
+    required: boolean;
+    scale: number;
+    searchPrefilterable: boolean;
+    sortable: boolean;
+    unique: boolean;
+    updateable: boolean;
+}
```
haifeng-li-at-salesforce, line 35: FieldInfo from `lds-lightning-platform`

lsp/server/src/types/ObjectInfoRepresentation.ts
```diff
@@ -0,0 +1,34 @@
...
+    mruEnabled: boolean;
+    nameFields: Array<string>;
+    queryable: boolean;
+    recordTypeInfos: {
+        [key: string]: any;
+    };
+    searchLayoutable: boolean;
+    searchable: boolean;
+    themeInfo: any;
+    updateable: boolean;
```
haifeng-li-at-salesforce, line 33: `ObjectInfo` type from `lds-lightning-platform`

lsp/server/src/utils/orgUtils.ts
```diff
@@ -0,0 +1,322 @@
...
+            return undefined;
+        }
+    }
+
+    // Updates the auth state async
+    private static async checkAuthStatus(): Promise<AuthStatus> {
+        if (this.authStatus !== AuthStatus.UNKNOWN) {
+            return this.authStatus;
+        }
+        const connection = await this.getConnection();
```
haifeng-li-at-salesforce, line 119: Update auth state if it is unknown

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { SFDX_PROJECT_FILE } from './constants';
+import * as fs from 'fs';
+import * as path from 'path';
+import { WorkspaceFolder } from 'vscode-languageserver';
```
haifeng-li-at-salesforce, line 11: WorkspaceFolder is different from the one in root project.

lsp/server/src/test/TestHelper.ts
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { Stats } from 'fs';
```
ben-zhang-at-salesforce, line 8: 8, 9 10 lines are unused imports

lsp/server/src/server.ts
```diff
@@ -153,6 +159,8 @@ connection.languages.diagnostics.on(async (params) => {
     }
 });
 
+// Watch SF config file change
+OrgUtils.watchConfig();
```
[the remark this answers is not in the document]
haifeng-li-at-salesforce: This is a good point. When a new workspace opens, the server is down and up again. So all the initialization will happen again. 
There is a potential memory link if previous server does not un-watch the fs. The un-watching is added.

lsp/server/src/test/TestHelper.ts
haifeng-li-at-salesforce: updated

lsp/server/src/types/FieldRepresentation.ts
ben-zhang-at-salesforce, line 1: it will be great if we  could import FieldRepresentation.ts and ObjectInfoRepresentation from npm sf published  instead of duplicate the physical file here.

lsp/server/src/types/ObjectInfoRepresentation.ts
ben-zhang-at-salesforce, line 1: same,  import from somewhere instead of duplicating?

lsp/server/src/utils/orgUtils.ts
```diff
@@ -0,0 +1,330 @@
...
+import { WorkspaceUtils } from './workspaceUtils';
+import { ObjectInfoRepresentation } from '../types';
+
+enum AuthStatus {
+    UNKNOWN,
+    AUTHORIZED,
+    UNAUTHORIZED
+}
+
+export interface Field {
```
ben-zhang-at-salesforce, line 27: is it used anywhere?

sfdctaka:
Members in Typescript class are public by default. Let's standardize by not explicitly putting the access modifier. Or, I'm fine doing the other way around too, by marking explicitly public.

lsp/server/src/utils/orgUtils.ts
```diff
@@ -0,0 +1,330 @@
...
+            this.connection = undefined;
+            return undefined;
+        }
+    }
+
+    private static async getEntityList(
+        connection: Connection
+    ): Promise<string[]> {
+        const globalResult = await connection.describeGlobal();
+        return globalResult.sobjects.map((sobjettResult) => sobjettResult.name);
```
sfdctaka, line 185: sobjett -> sobject

src/utils/constants.ts
```diff
@@ -4,10 +4,9 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-
+import { CORE_EXTENSION_ID, SFDX_PROJECT_FILE } from 'mobile-lsp-server';
```
sfdctaka, line 7: I think you can do 
`export  { CORE_EXTENSION_ID, SFDX_PROJECT_FILE } from 'mobile-lsp-server';`
Can you give it a try?

lsp/server/src/types/FieldRepresentation.ts
ben-zhang-at-salesforce: digged a bit and seems we have to do this - salesforce core doesn't have it at all.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: Good catch!

src/utils/constants.ts
haifeng-li-at-salesforce: Good point! Change to below way since `SETTING_SECTION_DIAGNOSTIC ` still needs import
import { CORE_EXTENSION_ID } from 'mobile-lsp-server';
export const SETTING_SECTION_DIAGNOSTIC = `${CORE_EXTENSION_ID}.diagnostics`;

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: Good catch! Removed

lsp/server/src/types/FieldRepresentation.ts
haifeng-li-at-salesforce: It is a trimmed version  from `lds-lightning-platform`, which is not publicly accessible.

haifeng-li-at-salesforce:
> Members in Typescript class are public by default. Let's standardize by not explicitly putting the access modifier. Or, I'm fine doing the other way around too, by marking explicitly public.

Good catch.  public members have been removed access modifier.

lsp/server/src/test/TestHelper.ts
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
```
sfdctaka, line 2: Copyright year

lsp/server/src/utils/orgUtils.ts
```diff
@@ -0,0 +1,327 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
```
sfdctaka, line 2: Copyright year.

lsp/server/src/utils/orgUtils.ts
```diff
@@ -0,0 +1,327 @@
...
+
+        await aggregator.reload();
+
+        const currentUserConfig = aggregator.getInfo(
+            OrgConfigProperties.TARGET_ORG
+        );
+
+        if (currentUserConfig.value) {
+            this.orgName = currentUserConfig.value.toString();
+            return Promise.resolve(this.orgName);
```
sfdctaka, line 76: You can just return `this.orgName`. It will automatically be wrapped in a promise.

lsp/server/src/utils/orgUtils.ts
```diff
@@ -0,0 +1,327 @@
...
+        }
+        return Promise.reject('no org');
```
sfdctaka, line 78: You can throw error with `throw new Error('no org')`. This has benefit of passing stack trace to the caller.

lsp/server/src/test/TestHelper.ts
ben-zhang-at-salesforce: @sfdctaka really impressed by your meticulous attentions to detail.

src/utils/constants.ts
```diff
@@ -4,10 +4,9 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-
+import { CORE_EXTENSION_ID } from 'mobile-lsp-server';
```
sfdctaka, line 7: `SETTING_SECTION_DIAGNOSTIC` in line 13 is not used anywhere in the project. We can remove line 7, right?

lsp/server/src/index.ts
sfdctaka: Can we improve on this? We are importing a module used to test into production code. If `TempProjectManager` is needed to be used in both for production and test then it should not be in `TestHelper`.

lsp/server/src/test/TestHelper.ts
sfdctaka: This class should not be in the test helper. It's being accessed in the production code. So it is a first class citizen in the codebase.

What's yucky is the way it is setup you need to reexport this class from root's index.ts. Let's move it to a place because it deserves it. Then we can get rid of reexport.

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    static isSfdxProjectOpened(): boolean {
+        try {
+            return fs.existsSync(
+                path.join(this.getWorkspaceDir(), SFDX_PROJECT_FILE)
+            );
+        } catch {
+            return false;
+        }
+    }
+}
```
sfdctaka, line 62: These two methods should be mashed.

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { SFDX_PROJECT_FILE } from './constants';
+import * as fs from 'fs';
+import * as path from 'path';
+import { WorkspaceFolder } from 'vscode-languageserver';
+
+export class WorkspaceUtils {
+    private static workspaceFolders: WorkspaceFolder[] | undefined | null = [];
```
sfdctaka, line 14: Remove `private`.

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    static readonly LWC_PATH = path.join(
+        WorkspaceUtils.DEFAULT_APP_PATH,
+        'lwc'
+    );
+
+    public static setWorkSpaceFolders(
+        workSpaceFolders: WorkspaceFolder[] | null | undefined
+    ) {
+        this.workspaceFolders = workSpaceFolders;
+    }
```
sfdctaka, line 31: `workSpaceFolders` shouldn't be undefined but either an array of `WorkspaceFolder` or `null`. Let's get rid of `undefined` here and in line 14.

```
export interface WorkspaceFoldersInitializeParams {
    /**
     * The workspace folders configured in the client when the server starts.
     *
     * This property is only available if the client supports workspace folders.
     * It can be `null` if the client supports workspace folders but none are
     * configured.
     *
     * @since 3.6.0
     */
    workspaceFolders?: WorkspaceFolder[] | null;
}
```

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    // LSP server has a different way to fetch WorkSpaceDir than root project which relis on vscode workspace.
```
sfdctaka, line 33: relies

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    static getWorkspaceDir(): string {
```
sfdctaka, line 34: This method gets created only inside this file. It throws error when there is no workspace. But the caller of this method catches that error and returns false. Which means, this error is serving no purpose. 

There are number of ways to clean up this code. I'll list one here.

You can get rid of `getWorkspaceDir()`. And assuming that you're mashing `lwcFolderExists()` and `isSfdxProjectOpened()` into one function, then, before you call `fs.existsSync()` in there you can do the same conditional
```
if (!this.workspaceFolders || this.workspaceFolders.length === 0) {
```
and see if you can do early exit if no valid workspace folders have been set yet.

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+export class NoWorkspaceError extends Error {
+    constructor(message?: string) {
+        super(message);
+        this.name = this.constructor.name;
+        Object.setPrototypeOf(this, NoWorkspaceError.prototype);
+    }
+}
```
sfdctaka, line 70: Unless you're logging this error or bubbling up this error to the caller, which is currently ending up being contained inside the `try...catch`, we can get rid of error class.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: This is not an error if default org does not exist

lsp/server/src/test/TestHelper.ts
```diff
@@ -0,0 +1,153 @@
...
+            try {
+                const projectDir = await mkdtemp(
+                    path.join(os.tmpdir(), 'offlineWizard-')
+                );
+                return resolve(new TempProjectDirManager(projectDir));
+            } catch (err) {
+                return reject(err);
+            }
+        });
+    }
```
sfdctaka, line 70:
```suggestion
        const projectDir = await mkdtemp(path.join(os.tmpdir(), 'offlineWizard-'));
        return new TempProjectDirManager(projectDir);
```

sfdctaka: Return a value from a function that returns promise should automatically wrap the return value.

For error, you're re-throwing the error in the form `return reject(err)` anyway so might as well just run the code and let it throw on its own and let the caller just handle it.

src/utils/constants.ts
haifeng-li-at-salesforce: Good catch! our code becomes cleaner

lsp/server/src/utils/orgUtils.ts
sfdctaka: It's not useful because `reject` will go into caller's `catch`, in `getDefaultUserName()`.  There you are not using `'no org'` in any way. All you are doing is catch an error and return `undefined`. If you throw error like this

`throw new Error('no org')` that will still land in the `catch` block of the same function. And then in there you can access `'no org'` in the form of `error.message` should you decide to log it.

PR #146, opened 2024-10-28
Implement over size field and over size record diagnostic for graphql query

ben-zhang-at-salesforce:
Generate lsp diagnostic if the graphql query fetch 
- individual text field whose length is over 32k.
- sum of the fetched field length is over 32 for an entity type.


lsp/server/src/validateGraphql.ts
```diff
@@ -10,14 +10,14 @@ import { gqlPluckFromCodeStringSync } from '@graphql-tools/graphql-tag-pluck';
 import { Diagnostic } from 'vscode-languageserver/node';
 import { DiagnosticProducer } from './diagnostic/DiagnosticProducer';
 import { TextDocument } from 'vscode-languageserver-textdocument';
-import { MisspelledUiapi } from './diagnostic/gql/misspelled-uiapi';
 import {
     DiagnosticSettings,
     isTheDiagnosticSuppressed
 } from './diagnostic/DiagnosticSettings';
+import { OversizedRecord as OversizedRequest } from './diagnostic/gql/over-sized-record';
```
sfdctaka, line 17: Do you need to `as` here?

lsp/server/src/utils/workspaceUtils.ts
```diff
@@ -24,47 +25,43 @@ export class WorkspaceUtils {
...
+    /**
+     * WorkspaceFolders must be initialized before they can be used.
+     * @param workspaceFolders workspaceFolders passed from LSP server connection.
+     */
+    public static initWorkspaceFolders(
+        workspaceFolders: WorkspaceFolder[] | null | undefined
     ) {
-        this.workspaceFolders = workSpaceFolders;
+        if (workspaceFolders === undefined) {
+            throw new Error('LSP client does no support workspace folder');
```
sfdctaka, line 36: Should we set `this.workspaceFolder` to `undefined` before throwing the error?

Also, is it possible that we not use undefined? I mean we know that `this.workspaceFolders` is of type `WorkspaceFolder[]`. So can it not just be `null`? `undefined` is for variable without a known type or without a value. I understand that the original code was written as `WorkspaceFolder[] | undefined | null = [];`. But you made the effort to removed `undefined` in line 15 so we should try to get rid of `undefined`, right?

Below is declaration for `WorkspaceFoldersInitializeParams`. Looks like `workspaceFolders` will never be `undefined`. We call `initWorkspaceFolders(workspaceFolders)` in line 49 of `server.ts` only and this looks like workspace folders will never be `undefined`.

image

lsp/server/src/utils/orgUtils.ts
```diff
@@ -274,17 +242,20 @@ export class OrgUtils {
                         return resolve(objectInfo);
                     }
                 } catch (e) {
-                    return resolve(undefined);
+                    console.log(
+                        `Failed to load entity list from server with error: ${e}`
+                    );
                 }
+                return resolve(undefined);
```
sfdctaka, line 249: Could this resolve go back inside the catch block?

lsp/server/src/utils/workspaceUtils.ts
ben-zhang-at-salesforce: good catch, let me fix it.

lsp/server/src/utils/orgUtils.ts
```diff
@@ -116,18 +76,26 @@ export class OrgUtils {
...
+    // Update the auth state async
     private static async checkAuthStatus(): Promise<AuthStatus> {
         if (this.authStatus !== AuthStatus.UNKNOWN) {
             return this.authStatus;
         }
+        try {
+            this.orgName = await this.getDefaultOrg();
+        } catch (e) {
+            this.orgName = undefined;
+        }
```
sfdctaka, line 88: This looks funny. What if in line 85 you get a valid org name back. And let's say the code goes into the `if` block. Then in line 92 why are you setting it to `undefined`? Another way to ask this question is do you need to preeempt to get org name in line 85? Previous version of this file wasn't doing it. I like to understand the significance.

lsp/server/src/utils/orgUtils.ts
```diff
@@ -116,18 +76,26 @@ export class OrgUtils {
...
         if (connection === undefined) {
             //It is possible that orgName exists and connection expires
-            this.orgName = '';
+            this.orgName = undefined;
             this.authStatus = AuthStatus.UNAUTHORIZED;
         } else {
             this.authStatus = AuthStatus.AUTHORIZED;
+            if (this.orgName === undefined) {
+                throw new Error('Authorized Org does not have org name!');
+            }
```
sfdctaka, line 98: Is it possible to have `undefined` org name when the auth status is `AUTHORIZED`?

lsp/server/src/utils/orgUtils.ts
```diff
@@ -311,17 +282,15 @@ export class OrgUtils {
...
-            } catch (e) {
-                console.log(e);
-            }
-            this.orgName = '';
+        try {
+            fs.rmSync(this.sfMobileFolder, {
+                force: true,
+                recursive: true,
+                maxRetries: 3
+            });
```
sfdctaka, line 290: What happens when you're logged into multiple orgs?

lsp/server/src/utils/workspaceUtils.ts
ben-zhang-at-salesforce: looked a bit deeper. seems the workspaceFolders could be undefined. 
image

let me think a bit more what do if client doesn't support workspace folders at all ( workspaceFolder will be undefined ).

lsp/server/src/server.ts
```diff
@@ -129,6 +128,21 @@ connection.onDidChangeConfiguration((change) => {
...
+const MAX_WAIT_FOR_STATE_AGGREGATOR = 4000;
+connection.onDidChangeWatchedFiles((changeEvents) => {
+    changeEvents.changes.forEach((change) => {
+        /**
+        When the default organization changes, the target_id in config.json will be updated. 
+        To handle this file change, we invoke onAuthOrgChanged. 
+        We've noticed that the StateAggregator in the Salesforce code may take over 3 seconds to stabilize, so 
+        we've implemented a fixed delay of up to 4 seconds here.  
+        */
+        if (change.uri.endsWith('.sf/config.json')) {
```
sfdctaka, line 140: What happens if the user is still using `sfdx` instead of `sf`?

lsp/server/src/utils/gqlUtils.ts
```diff
@@ -0,0 +1,420 @@
...
+import { ObjectInfoRepresentation } from '../types';
+
+/**
+ * Represent an entity in graphql query.
+ */
+export interface EntityNode {
+    // Graphql AST node
+    node: FieldNode;
+    // Entity name, like `Account`
+    name: string | undefined;
```
sfdctaka, line 21: Can there be a node without a name?

lsp/server/src/utils/gqlUtils.ts
```diff
@@ -0,0 +1,420 @@
...
+    // Total size of immediate children fields
+    size: number | undefined;
```
sfdctaka, line 23: Suppose there was no children. Then, shouldn't `size` suffice to be only `number`? Can it also be `undefined`?

lsp/server/src/utils/workspaceUtils.ts
haifeng-li-at-salesforce: 'undefined' means that 'workspace' is not supported, which actually would fails all the tests we have. So it is filtered out immediately inside 'initWorkspaceFolder'

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: If 'connection' is 'undefined', it will never resolve if 'resolve' is put into 'catch'

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: 'OrgName' is used to construct 'ObjectInfoFolder'. If  connection is invalid, its value should be set to 'undefined' to avoid dirty state.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: Not possible. This is the safeguard which correlates to the OrgName change

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: Every time user log into a new org, the cache under '.sfMobile' is cleared. It behaves the same when user switch org

lsp/server/src/server.ts
haifeng-li-at-salesforce: This is a very good question. I think we can enforce the minimum required core extension version which is inside 'constants.ts'

lsp/server/src/utils/gqlUtils.ts
haifeng-li-at-salesforce: 'undefined' means it is unknown yet. When 'EntityNode' is constructed in 'EntityTree', it is possible that both 'name' and 'size' are unknown. These properties will become available when traversal happens and objectInfo is fetching .

lsp/server/src/server.ts
haifeng-li-at-salesforce: We can also monitor `config.json' under '.sfdx'

lsp/server/src/diagnostic/gql/over-sized-record.ts
```diff
@@ -0,0 +1,167 @@
...
+    for (const operationNode of rootNode.operations) {
+        for (const entityNode of operationNode.entities) {
+            try {
+                await generateDiagnostic(entityNode, results);
+            } catch (e) {
+                console.log(
+                    `Cannot conduct over-sized record diagnostic for ${entityNode.name} `
+                );
+            }
+        }
```
sfdctaka, line 85: Suppose `generateDiagnostic` throws where `results` from previous iteration has already accumulated elements. Then in this case is it an expected behavior that `results` may contain diagnostics? Or should this function be returning empty array?

lsp/server/src/validateGraphql.ts
ben-zhang-at-salesforce: removed

lsp/server/src/utils/orgUtils.ts
sfdctaka: This `connection` is not `undefined` then you should early exit after 
```
const connection = await OrgUtils.getConnection();
if (connection === undefined) {
  return resolve(objectInfo);
}
```
You don't return based on thing that you do in a `try` block.

Without you having telling me I would not have known.

Besides, 
```
return resolve(undefined);
```
was previously in the `catch` block.

lsp/server/src/utils/workspaceUtils.ts
haifeng-li-at-salesforce: I don't encounter this situation in my environment, but it is possible in other user's environment.

lsp/server/src/utils/orgUtils.ts
sfdctaka: But org name being used to construct `ObjectInfoFolder` has no relevance here.

Compared your change to the previous version. Org name must have been fetched and assigned to `this.orgName` else where. In your code, instead, you decided to do it in `checkAuthStatus()`. Why?

And I go back to my question again. Suppose there was valid `orgName`. But in line 92 if you don't have connection you set `orgName` to undefined. Line 92 is almost identical to the previous version of the code. But I think that one was wrong too for this part. `CheckAuthStatus()` should only check and not change states, such as `orgName`. This is where this function can be improved/refactored. In another words, `checkAuthStatus()` which is signaling for read, in fact, is also doing write.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: The 'resolve' is moved out of 'catch' so that it handle two conditions:

1.  Connection is undefined.
2. any error is thrown.
We can put one `resolve(undefined)` inside 'catch`, another check under 'const connection = await OrgUtils.getConnection();'. It is the same thing. It is not a late exit, if connection is undefined, it will go to line 249 immediately.

lsp/server/src/utils/orgUtils.ts
sfdctaka: If it is not possible why bother with having a safeguard?

lsp/server/src/server.ts
haifeng-li-at-salesforce: Watch  '.sfdx/sfdx-config.json' now. Code is updated.

lsp/server/src/utils/orgUtils.ts
sfdctaka: Isn't this more destructive? Previous version of the code was specified to remove the folder at object info folder path. Whereas, your code nukes at the `.sfMobile` level. Are we certain that we want to do this?

lsp/server/src/diagnostic/gql/over-sized-record.ts
haifeng-li-at-salesforce: Good question. Each EntityNode is an independent unit from its peers. Error in one Entity does not interfere with its peer. So iteration should continue if one entityNode processing produces error.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: `OrgName` and `authStatus` are the  org states. In previous version, they are maintained in 'getDefaultOrg' and `checkAuthStatus` separately.  In this version, `orgName` is pulled out from `getDefaultOrg` and put into `checkAuthStatus` as the central place to maintain status. `CheckAuthStatus`'s name might be misleading, but it does state maintaining job.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: In theory, it is not possible. The safeguard code is put there in case some bug comes in and we know what the problem is. I will take it out.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: Good question. It is more secure to me. When user logs/switches into a new org or log out, the cached data from previous org should be wiped out to prevent other people peeking into it. We can discuss this topic on the stand-up.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: The code is taken out

lsp/server/src/utils/orgUtils.ts
sfdctaka: If your argument stands then why wasn't it done your way previously?

lsp/server/src/utils/workspaceUtils.ts
sfdctaka: Here's the code
https://github.com/microsoft/vscode-languageserver-node/blob/ce4475fc88130e27665f49df49f162e20c328660/protocol/src/common/protocol.workspaceFolder.ts#L11

And here's the spec:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders

lsp/server/src/diagnostic/gql/over-sized-record.ts
sfdctaka: Can we also log `e.message` please.

lsp/server/src/diagnostic/gql/over-sized-record.ts
```diff
@@ -0,0 +1,167 @@
...
+                    continue;
+                }
+                relation.entity.name = entityName;
+            }
+
+            try {
+                await generateDiagnostic(relation.entity, overSizedDiagnostic);
+            } catch (e) {
+                console.log(
+                    `Cannot conduct over-sized record diagnostic for ${entityNode.name} `
```
sfdctaka, line 146: Same here, let's log error's message since we're logging the entity with problem already.

lsp/server/src/utils/orgUtils.ts
```diff
@@ -182,15 +148,15 @@ export class OrgUtils {
...
-    // Retrieves objectInfo folder path, which is '<projectRoot>/.sf/orgName/objectInfos/'
+    // Retrieve objectInfo folder path, which is '<projectRoot>/.sfMobile/orgName/objectInfos/'
     private static objectInfoFolderPath(): string {
         const projectPath = WorkspaceUtils.getWorkspaceDir();
         if (this.orgName === undefined || this.orgName.length === 0) {
             throw new Error('AuthError: No Org exists');
         }
         const objectInfoFolder = path.join(
             projectPath,
-            this.sfFolder,
```
sfdctaka, line 193: Now I'm not sure if `sfFolder` is correct. If the user is using `sfdx` instead of `sf` wouldn't things be saved under `.sfdx` folder instead of `.sf`?

The assumption made here that it is always `.sf` will return a biased path when `objectInfoFolderPath()` is called.

lsp/server/src/utils/orgUtils.ts
sfdctaka: Furthermore. 

Look at my comment on how `ojectInfoFolderPath()` is constructed. Now, I don't remember if anything can get written in `.sfdx` folder in the current setup. If it does, then, isn't it better to keep how you were removing the folders previously than just `sfMobileFolder`?

Also, I vaguely remember if we cache things for the wizard. Could nuking `.sf` folder have any ill effect?

lsp/server/src/utils/orgUtils.ts
sfdctaka: Consider this. Simplified try catch. No post processing of what had happened inside the try block.

```
try {
    const connection = await OrgUtils.getConnection();

    if (connection === undefined || !OrgUtils.entities.includes(objectApiName)) {
        return resolve(undefined);
    }

    const objectInfo = await connection.request(
        `${connection.baseUrl()}/ui-api/object-info/${objectApiName}`
    ) as ObjectInfoRepresentation;

    if (objectInfo !== undefined) {
        this.objectInfoResponseCallback(objectApiName, objectInfo);
    }

    return resolve(objectInfo);

} catch (error) {
    console.log(`Error when calling getObjectInfo: ${error}`);
    return resolve(undefined); // Return undefined in case of an error
}

```

lsp/server/src/utils/orgUtils.ts
sfdctaka: Why did you pull out `getDefaultOrg`? That seems to be a good place to keep the one and only place to manage `orgName` state?

lsp/server/src/utils/orgUtils.ts
```diff
@@ -116,18 +76,24 @@ export class OrgUtils {
...
         }
+        try {
+            this.orgName = await this.getDefaultOrg();
+        } catch (e) {
+            this.orgName = undefined;
+        }
         const connection = await this.getConnection();
         if (connection === undefined) {
             //It is possible that orgName exists and connection expires
-            this.orgName = '';
```
sfdctaka, line 127: This line from previous version of the code is evil too. It's trying to do a service by if the connection is not available then set org's name to empty string. Which, you emulated but instead assign an undefined. Do you know if this line is necessary here or in your PR's case `this.orgName = undefined;`. Ideally `checkAuthStatus` itself shouldn't change the state but solely focus on doing what it claims to do. Principle of single responsiblity.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: Good suggestion!

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: There are two reasons to pull out orgName setting from `getDefaultOrg`

1. Centralize the state update within `checkAuthStatus`.  We don't want to have two places to maintain states. What is more, everytime `getObjectInfo` is called, `checkAuthStatus` is invoked at the beginning, which makes sure that state is accurate.
2. `getDefaultOrg` should be only read instead of update. This is the same reasoning behind as you have for `CheckAuthStatus`

sfdctaka: I think you and I differ on point #2. I don't think `getDefaultOrg` can be only be read only because it is a computed property. And throughout the code `this.orgName` is used for a stashed value.

But you can make `getDefaultOrg` read only if you can get rid of all usages `this.orgName` and instead call `getDefaultOrg` whenever you need the org name. That way no stashed value ever goes stale so you don't have to worry about needing to update `this.orgName` in `checkAuthStatus`.

Is there a cost to not stashing? Brief look at `getDefaultOrg` doesn't look like there is one. Even if the aggregator is making file I/O the amount of data that needs to be read off the disk is minimal that I wouldn't consider it a performance hit.

lsp/server/src/utils/orgUtils.ts
sfdctaka: Got it. So .sfMobile is only for stashing object info. And watching .sf and .sfdx is for auth status?

lsp/server/src/diagnostic/gql/over-sized-record.ts
haifeng-li-at-salesforce: added

lsp/server/src/diagnostic/gql/over-sized-record.ts
haifeng-li-at-salesforce: added

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: We  use '.sfMobile' dir instead.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: `checkAuthStatus` is the central hub to maintain the state. The reason is listed in above comments.

lsp/server/src/utils/orgUtils.ts
haifeng-li-at-salesforce: yes

lsp/server/src/utils/orgUtils.ts
ben-zhang-at-salesforce: @sfdctaka centralized the org name, connection, status into new object OrgInfo , take a look and see if it address your concern.

lsp/server/src/utils/orgUtils.ts
ben-zhang-at-salesforce: fixed.

haifeng-li-at-salesforce (approved):
Very great after detailed review!

PR #147, opened 2024-10-30
Finalize draft not aware diagnostics for getRelatedListRecords and getRelatedListCount

ben-zhang-at-salesforce:
- Finalize the diagnostic UI for getRelatedListRecords and getRelatedListCount as below: 
image
- Skip LSP diagnostics if opened project is not sfdx or opened document is not under LWC folder.


lsp/server/src/diagnostic/js/adapters-local-change-not-aware.ts
```diff
@@ -45,22 +47,26 @@ export class AdaptersLocalChangeNotAware implements DiagnosticProducer<Node> {
                         start: textDocument.positionAt(item.start as number),
                         end: textDocument.positionAt(item.end as number)
                     },
-                    message: LOCAL_CHANGE_NOT_AWARE_MESSAGE
+                    message: LOCAL_CHANGE_NOT_AWARE_MESSAGE,
+                    code: LOCAL_CHANGE_NOT_AWARE_EXTERNAL_DOC_URL,
```
ben-zhang-at-salesforce, line 51: diagnostic renders code and codeDescription as link <a name = '$code$' href = '$codeDescription.href'>

lsp/server/src/server.ts
```diff
@@ -143,15 +143,16 @@ documents.onDidClose((e) => {
 });
 
 connection.languages.diagnostics.on(async (params) => {
-    const document = documents.get(params.textDocument.uri);
-    if (document !== undefined) {
+    const uri = params.textDocument.uri;
+    const document = documents.get(uri);
+
+    // Do diagnostics if document is under LWC folder and already in cache.
+    if (uri.indexOf(WorkspaceUtils.LWC_PATH) > 0 && document !== undefined) {
```
ben-zhang-at-salesforce, line 150: only do diagnostics for file under lwc folder

lsp/server/src/validateDocument.ts
```diff
@@ -45,10 +45,5 @@ export async function validateDocument(
         results = results.concat(diagnostics);
     }
 
-    // Set the source for diagnostic source.
-    results.forEach((diagnostic) => {
-        diagnostic.source = extensionName;
```
sfdctaka, line 50: Curious, why did you get rid of this? I wasn't a fan of this post processing anyway so removing it is a welcome change but where is the source set now?

ben-zhang-at-salesforce: The vscode has its own brain where and how to render the diagnostic attributes.  I like to render the diagnostic.source "Salesforce Mobile Extension in ui so to tell developer where this diagnostic is from. When we try to render a link in diagnostic pop up like what Kevin wants.  the source is rendered between the msg and cde/codeDescription like below. Totally messed up the effect we want.  Removing the source makes the popup close to what we want. 
image

Source will not be set if we will render the link. 
Kevin is thinking of what will be finally word for the rules, if there's no link needed.  we can put the source back.

sfdctaka: To make myself clear, I wasn't the fan of post processing, so removing is fine I think. But do what you think is the best thing to do.

ben-zhang-at-salesforce:
I messed up the change while rebase, will close this one and create a new pr.

PR #148, opened 2024-10-31
Validator pattern change

haifeng-li-at-salesforce:
### What does this PR do?
 consolidate validator functions into consistent object structure
### What issues does this PR fix or reference?

- Duplicate code across validators
- Rigid structure for managing validators, limiting flexibility
- Sequential execution instead of parallel execution
- Error should be contained within producers, preventing them from propagating unnecessarily.
- Hardcoded producer lists in each validator


lsp/server/src/validatorManager.ts
```diff
@@ -0,0 +1,162 @@
...
+     * @param document The document to validate.
+     * @param extensionName The name of the extension (sets diagnostic source).
+     * @returns A promise resolving to an array of diagnostics.
+     */
+    async validateDocument(
+        setting: DiagnosticSettings,
+        document: TextDocument,
+        extensionName: string
+    ): Promise<Diagnostic[]> {
+        const qualifiers = this.validators.filter(
```
ben-zhang-at-salesforce, line 49: nitpick:  something like below ```suggestion
        const validatorsNeedToRun = this.validators.filter(
```

File: lsp/server/src/validatorManager.ts [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+            (validator) => validator.getLanguageId() === document.languageId
+        );
+        const diagnosticArray = await Promise.all(
+            qualifiers.map(async (validator) => {
+                try {
```
ben-zhang-at-salesforce, line 54: great idea to try catch here to avoid exception in on validator effect other validator

ben-zhang-at-salesforce (approved):
looks good except one minor nitpicking.  great to centralize the error catch, two level parsing in a general way. great job!

PR #150, opened 2024-11-04
chore: upgrade to node >=20

maliroteh-sf:
### What does this PR do?
Upgrade to node >=20

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -14,7 +14,7 @@ jobs:
             - name: Install Node.js
               uses: actions/setup-node@v3
               with:
-                  node-version: '18.x'
+                  node-version: '20.x'
 
             - name: Install dependencies
               run: npm install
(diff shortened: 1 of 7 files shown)


sfdctaka:
Do we want to delete package-lock.json?

maliroteh-sf:
> Do we want to delete package-lock.json?

@sfdctaka This repo does not use `yarn` so did you mean whether we want to delete `package-lock.json` and then run `npm install` to regenerate the file with updated dep versions? We now have dependabot for such things so I didn't do such update in this PR and tried to keep it as simple as possible.

sfdctaka:
I made a mistake. I thought github had package-lock.json as deleted. Instead, the file is too big it didn't show the diff by default.

khawkins:
Or...do you think we deleted it and that's the question? 😀 FWIW I don't see anything amiss with `package-lock.json` in this PR.

sfdctaka (approved):
LGTM!

PR #153, opened 2024-11-09
Finalize draft not aware Finalize draft not aware diagnostics for getRelatedListRecords and getRelatedListCount

ben-zhang-at-salesforce:
getRelatedListRecords is finalized to 
image

getRelatedListCount is finalized to 
image

### DIFF
--- a/lsp/client/src/extension.ts
+++ b/lsp/client/src/extension.ts
@@ -37,10 +37,6 @@ export function activate(
         }
     };
 
-    // Get extension name
-    const extensionTitle =
-        context.extension.packageJSON.contributes.configuration.title;
-
     // Options to control the language client
     const clientOptions: LanguageClientOptions = {
         documentSelector: [
@@ -57,7 +53,6 @@ export function activate(
             ]
         },
         initializationOptions: {
-            extensionTitle,
             updateDiagnosticsSettingCommand,
             diagnosticsSettingSection
         }
--- a/lsp/server/src/diagnostic/js/adapters-local-change-not-aware.ts
+++ b/lsp/server/src/diagnostic/js/adapters-local-change-not-aware.ts
@@ -53,9 +53,9 @@ export class AdaptersLocalChangeNotAware implements DiagnosticProducer<Node> {
     ): Promise<Diagnostic[]> {
         const result: Diagnostic[] = [];
 
-        const nodesWithAdapter = this.findLocalChangeNotAwareAdapterNode(node);
+        const nodeInfoMap = this.findLocalChangeNotAwareAdapterNode(node);
 
-        for (let [node, info] of nodesWithAdapter) {
+        for (let [node, info] of nodeInfoMap) {
             const diagnostic: Diagnostic = {
                 severity: SEVERITY,
                 range: {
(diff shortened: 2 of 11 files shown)


lsp/client/src/extension.ts
```diff
@@ -37,10 +37,6 @@ export function activate(
         }
     };
 
-    // Get extension name
```
ben-zhang-at-salesforce, line 40: have to remove the diagnostic.source which has extension title as it's value due to the popup renders it between the msg and code(link).

lsp/server/src/validator/baseValidator.ts
```diff
@@ -90,7 +90,15 @@ export abstract class BaseValidator<SupportedType> {
...
             activeProducers.map(async (producer) => {
                 try {
-                    return await producer.validateDocument(textDocument, data);
+                    const producerId = producer.getId();
+                    const diagnostics = await producer.validateDocument(
+                        textDocument,
+                        data
+                    );
+                    diagnostics.forEach((diagnostic) => {
+                        diagnostic.data = producerId;
```
ben-zhang-at-salesforce, line 99: need to add the producerId otherwise the suppressing will not work.

lsp/client/src/extension.ts
haifeng-li-at-salesforce: This deletion makes code cleaner.  Good move!

lsp/server/src/diagnostic/js/adapters-local-change-not-aware.ts
```diff
@@ -34,39 +51,43 @@ export class AdaptersLocalChangeNotAware implements DiagnosticProducer<Node> {
...
-                        start: textDocument.positionAt(item.start as number),
-                        end: textDocument.positionAt(item.end as number)
-                    },
-                    message: LOCAL_CHANGE_NOT_AWARE_MESSAGE
-                } as Diagnostic;
-            })
-        );
+        const result: Diagnostic[] = [];
+
+        const nodesWithAdapter = this.findLocalChangeNotAwareAdapterNode(node);
```
haifeng-li-at-salesforce, line 56: Is the `nodeInfoMap` a better name?

lsp/server/src/diagnostic/js/adapters-local-change-not-aware.ts
```diff
@@ -34,39 +51,43 @@ export class AdaptersLocalChangeNotAware implements DiagnosticProducer<Node> {
...
         export default class RelatedListRecords extends LightningElement {
             ...
             @wire(getRelatedListRecords, 
             ...
         }
-     * @param astNode root node to search
-     * @param adapterNames adapter which are not able to reflect the local change.
-     * @returns nodes with adapter name
+     * @param astNode The AST root node for the LWC js file.
+     * @returns A map of node to adapter name which is not draft aware.
```
haifeng-li-at-salesforce, line 87: How about `A map where each key is an ASTNode representing an adapter that does not account for specific local changes, and each value is an Info object containing a message and, optionally, a link with additional context.`

haifeng-li-at-salesforce:
Look great except a few naming and documentation suggestion.

haifeng-li-at-salesforce (approved):
Change is great.

PR #155, opened 2024-11-12
Move test folders

sfdctaka:
### What does this PR do?

Moved test folders for each package. Making them sit at the level as respective `src` folders.


lsp/server/tsconfig.json
```diff
@@ -9,10 +9,9 @@
         "sourceMap": true,
         "strict": true,
         "outDir": "out",
-        "rootDir": "src",
+        "rootDir": ".",
         "composite": true
     },
-
-    "include": ["src"],
+    "include": ["src", "test", "testFixture/**/*.json"],
```
sfdctaka, line 15: Test fixture needs to be specified in `include`, otherwise can't be found.

haifeng-li-at-salesforce:
The code functions as expected, but I have some concerns about the structure of the compiled code.

The main goal of relocating the test folder was to prevent it from being included in the vsix package. However, after moving test to the same directory level as src, the test folder still ends up generated and included in the compiled out directory. Does this outcome go against our original intention?

Is there a way to run all the tests without having them included in the `out` folder, similar to to the approach used in https://github.com/salesforce/eslint-plugin-lwc-mobile ?

haifeng-li-at-salesforce (approved):
The vsce has a way to exclude folders. Changes looks great to me. Good job!

package.json
```diff
@@ -95,14 +95,15 @@
...
+        "copy-resources": "cp -r './lsp/server/resources' './lsp/server/out/src/'",
         "watch": "tsc -watch -b ./",
         "pretest": "npm run compile && npm run lint",
         "lint": "eslint src --ext ts",
         "test": "node ./out/test/runTest.js",
         "test-coverage": "node ./out/test/runTest.js --coverage",
         "prettier:write": "prettier --write \"src/**/*.{ts, js}\" \"resources/instructions/**/*.html\" \"lsp/server/src/**/*.{ts, js}\" \"lsp/client/src/**/*.{ts, js}\"",
-        "prettier:verify": "prettier --list-different \"src/**/*.{ts, js}\" \"resources/instructions/**/*.html\" \"lsp/server/src/**/*.{ts, js}\" \"lsp/client/src/**/*.{ts, js}\"",
+        "prettier:verify": "prettier --list-different \"src/**/*.{ts, js}\" \"resources/instructions/**/*.html\" \"lsp/server/src/**/*.{ts, js}\" \"lsp/client/src/**/*.{ts, js}\""
+        ,
```
maliroteh-sf, line 106: Is it intentional to put the `,` on the next line (maybe prettier did it)?

maliroteh-sf (approved):

ben-zhang-at-salesforce (approved):
great job! thx for moving the tests out of src

PR #157, opened 2024-11-15
Fix packaging

sfdctaka:
### What does this PR do?
This PR consolidates 3 packages into one. There were 3 packages:
1. Root. The gateway to extension
2. LSP client
3. LSP server

Consolidation allows:
1. Packaging. **npm workspace** which was used before is removed because it was using symbolic links to child packages. And `npx vsce package` wasn't supporting symbolic links thus not producing `vsix` distributable.
2. Give us a clean slate to eventually break the single package into multiple packages. Before this PR a clean code sharing was hard and code duplication was made.


src/commands/settings/settings.ts
```diff
@@ -6,15 +6,14 @@
...
 import * as vscode from 'vscode';
-import { getExtensionName } from '../../utils/workspaceUtils';
 
 const SETTING_KEY_SUPPRESS_ALL = 'suppressAll';
 const SETTING_KEY_SUPPRESS_BY_RULE_ID = 'suppressByRuleId';
 
 export function getUpdateDiagnosticsSettingCommand(
     context: vscode.ExtensionContext
 ): string {
-    return `${getExtensionName(context)}.updateDiagnosticsSetting`;
```
sfdctaka, line 17: `getExtensionName()` was only used here so no need to have a method defined in `workspaceUtils.ts`.

src/lsp/client/client.ts
```diff
@@ -60,8 +60,8 @@ export function activate(
 
     // Create the language client and start the client.
     client = new LanguageClient(
-        'LSP Mobile',
-        'LSP Mobile Client',
+        'Salesforce LSP Mobile',
+        'Salesforce LSP Mobile Client',
```
sfdctaka, line 64: Updated name so it is displayed as a unique entry in VSCode.
image

src/lsp/server/utils/objectInfoUtils.ts
```diff
@@ -75,11 +75,11 @@ class OrgInfo {
 }
 
 /**
- * Utility class for Org
+ * Utility class for managing ObjectInfo
  * - fetching data and cache it in L1 and L2 for authorized org
  * - wipe out data when logout.
  */
-export class OrgUtils {
+export class ObjectInfoUtils {
```
sfdctaka, line 82: Renamed the class to reflect what it manages instead of where it came from.

src/lsp/server/utils/serverWorkspace.ts
```diff
@@ -5,35 +5,20 @@
...
  */
 
-import { SFDX_PROJECT_FILE } from './constants';
-import * as fs from 'fs';
-import * as path from 'path';
 import { WorkspaceFolder } from 'vscode-languageserver';
 
-export class WorkspaceUtils {
-    // WorkspaceFolders is null if Workspace is not configured
+export class ServerWorkspace {
```
sfdctaka, line 10: Much of `ServerWorkspace` came from `WorkspaceUtils` which most of it was a code from previous iteration prior to LSP. The reason LSP needs a `ServerWorkspace` was because it wanted to stash `workspaceFolders` property(next line in the code) that gets passed to when the LSP server is initialized.

Rather than duplicating `WorkspaceUtils`, I created a property bag class with managing `workspaceFolders` property. Static methods are essentially the same as they were implemented when LSP feature was done.

src/utils/constants.ts
```diff
@@ -5,6 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 export const MINIMUM_REQUIRED_VERSION_CORE_EXTENSION = '58.4.1';
-export { CORE_EXTENSION_ID, SFDX_PROJECT_FILE } from 'mobile-lsp-server';
+export const CORE_EXTENSION_ID = 'salesforce.salesforcedx-vscode-core';
+export const SFDX_PROJECT_FILE = 'sfdx-project.json';
```
sfdctaka, line 9: These constants, too, were originally defined in root, which moved to LSP server and then reverse exported back to the root project. Since we're consolidating they're back to where they were.

test/TestHelper.ts
```diff
@@ -5,12 +5,72 @@
...
+const testOrgConfig = {
+    key: 'target-org',
+    value: 'testOrg'
+};
+
+const noOrgConfig = {
+    key: 'target-org'
+};
+
+export class TempProjectDirManager {
```
sfdctaka, line 34: `TempProjectDirManager` coming back to the file where it was originally.

test/suite/lsp/server/diagnostics/gql/over-sized-record.test.ts
```diff
@@ -16,22 +16,23 @@ import {
...
 
-import { ObjectInfoRepresentation } from '../../../src/types';
+import { ObjectInfoRepresentation } from '../../../../../../src/lsp/server/types';
 
 suite(
     'GraphQL Diagnostics Test Suite - Server - Oversized GraphQL Field',
     () => {
+        let sandbox: sinon.SinonSandbox;
         let oversizedRecordProducer = new OversizedRecord();
+        const book = JSON.parse(readFileSync('test/suite/lsp/server/testFixture/objectInfos/Book__c.json', 'utf-8'));
```
sfdctaka, line 29: `resolveJsonModule` is no longer defined in `tsconfig`. So I'm reading json using `fs` instead.

src/lsp/client/client.ts
```diff
@@ -24,7 +24,7 @@ export function activate(
 ) {
```
sfdctaka, line 24: This file is renamed from `extension.ts` to `client.ts`. `Extension.ts` should be used for the main entry point in an extension. So following the same spirit as `lsp/server` I named it `client.ts`. I thought of changing `activate()` too since it could be anything if it is not an extension. Besides, it's taking extra parameters in `updateDiagnosticSettingCommand` and `diagnosticSettingSection` that aren't expected by extension loader. But changing the file name suffices so I left `activate()` as is.

src/lsp/server/utils/objectInfoUtils.ts
maliroteh-sf: Renaming the class makes sense but not the file. The file contains more than one class, which perform org-related tasks as a whole. So I think that we should have the file as OrgUtils still.

Or break out the 2 classes into their own files, especially since `ObjectInfoUtils` class is rather large?

src/lsp/server/utils/serverWorkspace.ts
maliroteh-sf: Hmm.... I'm not sure how I feel about `ServerWorkspace` as the name. This file was containing helper code around the concept of Workspace in an extension, and to that end `Server` or `Client` workspace doesn't make sense (not to me at least).... but I'll differ to others on this one.

src/lsp/server/utils/serverWorkspace.ts
```diff
@@ -5,35 +5,20 @@
...
      * WorkspaceFolders must be initialized before they can be used.
      * @param workspaceFolders workspaceFolders passed from LSP server connection.
      */
-    public static initWorkspaceFolders(
+    static initWorkspaceFolders(
         workspaceFolders: WorkspaceFolder[] | null | undefined
     ) {
         if (workspaceFolders === undefined) {
-            throw new Error('LSP client does no support workspace folder');
+            throw new Error('LSP client requires workspace folder.');
```
maliroteh-sf, line 21: Another reason why the name `ServerWorkspace` may not be the best option..... The class/file name is `ServerWorkspace` but the error message references `client`

sfdctaka: @haifeng-li-at-salesforce might be able to answer this better as I'm leaving much of code written originally for LSP's first implementation. If I were to attempt answering it, this method is ONLY being called in LSP server as it is getting bootstrapped. That bootstrap process is started by client. And that's why it probably has reference to `client`.

maliroteh-sf (commented):
Left some comments for your considerations. For some of the feedback we can see how the rest of the team feels about them so don't feel obligated to change things right away... if others are ok with the way things are in your PR then we can go as is.

src/lsp/client/client.ts
ben-zhang-at-salesforce: nitpick:   should we name it to 'Salesforce LSP mobile'? I make lsp server output some log like 
image
it's in the Salesforce LSP Mobile Client console stream.

tsconfig.json
```diff
@@ -12,10 +12,5 @@
         // "noFallthroughCasesInSwitch": true, /* Report errors for fallthrough cases in switch statement. */
         // "noUnusedParameters": true,  /* Report errors on unused parameters. */
     },
-
-    "include": ["src", "test"],
-    "references": [
-        { "path": "./lsp/client" }, 
-        { "path": "./lsp/server" }
-    ]
+    "include": ["src", "test", "test/suite/lsp/server/testFixture/**/*.json"]
```
khawkins, line 15: Can we set `rootDir` back to `src` as well? Lost that change in https://github.com/salesforce/salesforcedx-vscode-mobile/pull/155, and it's why we now have an `out/src` folder structure in our build.

src/lsp/client/client.ts
sfdctaka: @ben-zhang-at-salesforce Double checking. Your screenshot is chopped off but that's where you want the name to be "Salesforce LSP mobile", right?

.vscodeignore
```diff
@@ -13,6 +12,6 @@ RELEASE.md
 coverage/
 .nyc_output/
 .github
-out/
-!out/extension.js
-**/test/**
\ No newline at end of file
+!out/src
```
khawkins, line 15: You don't really need this once `out/` is gone from `.vscodeignore`.

tsconfig.json
sfdctaka: Yes, it was done intentionally because `test` folder now sits at the same level with `src`. I wanted to have these two folders reflect their same positions under `out` as well so that I can run tests and then to explicitly not include the test folder in vsix package(through .vscodeignore).

src/lsp/client/client.ts
haifeng-li-at-salesforce: Maybe `Salesforce Mobile Language Server` is a better fit ?

package.json
```diff
@@ -3,10 +3,7 @@
     "displayName": "%extension.displayName%",
     "description": "%extension.description%",
     "version": "0.3.0",
-    "publisher": "salesforce",
-    "workspaces": [
-        "lsp/*"
-    ],
+    "publisher": "Salesforce",
```
khawkins, line 6: This can't be changed. It's a specific value that the VSCode extension publishing facilities look for to associate the package, similar to an NPM package scope.

```suggestion
    "publisher": "salesforce",
```

src/lsp/server/utils/objectInfoUtils.ts
haifeng-li-at-salesforce: There is no need to rename the files or class names. OrgUtils encompasses more functionality than just fetching ObjectInfo. The ObjectInfo functions can be viewed as a subset of OrgUtils' capabilities.

khawkins:
Nit: Do we need a `yaml/` subdirectory under `resources/`? Seems like a folder structure that would benefit from being as flat as possible, at least for this project.

sfdctaka:
@khawkins `yaml/` folder is not necessary. I have it so that it would look like it's logically divided like its sibling folders `instructions` and `templates`. I'll remove it.

khawkins:
We lost `esbuild` as part of https://github.com/salesforce/salesforcedx-vscode-mobile/pull/127. We can't package without it.

src/lsp/server/utils/objectInfoUtils.ts
sfdctaka: Just another reminder that there is a class call OrgUtils in utils folder of the root project too that is absolutely unrelated. I will revert the names. It's strange, to me, that there are two separate OrgUtils.

test/suite/lsp/server/diagnostics/gql/over-sized-record.test.ts
sfdctaka: Actually resurrected `resolveJsonModule` and removed reading json using `fs`. I changed how the code was `import`ing JSONs to make them work.

ben-zhang-at-salesforce (approved):
looks good for me, great job.

src/lsp/server/utils/serverWorkspace.ts
maliroteh-sf: Discussed offline and will keep as is in this PR

haifeng-li-at-salesforce: WorkSpaceFolder in the LSP server and VS Code are two distinct classes. In the previous workspace structure, the server and client were independent packages, so using workspace as a name made sense within a self-contained project. Now that we've consolidated everything into a single package, renaming it to ServerWorkSpace is more appropriate to distinguish it from VS Code's workspace.

package.json
```diff
@@ -93,20 +90,27 @@
...
     },
     "scripts": {
-        "clean": "rimraf out ./lsp/server/out ./lsp/client/out",
-        "compile": "rimraf ./lsp/server/tsconfig.tsbuildinfo ./lsp/client/tsconfig.tsbuildinfo && tsc -b ./ && npm run copy-resources",
-        "copy-resources": "cp -r './lsp/server/resources' './lsp/server/out/src/'",
+        "clean": "rimraf ./out",
+        "compile": "rimraf ./out/tsconfig.tsbuildinfo && tsc -b ./ && npm run copy-yaml",
+        "copy-yaml-win": "mkdir out\\src\\resources && copy resources\\component-experiences.yaml out\\src\\resources",
+        "copy-yaml-unix": "mkdir -p ./out/src/resources && cp ./resources/component-experiences.yaml ./out/src/resources",
+        "copy-yaml": "npm run copy-yaml-win || npm run copy-yaml-unix",
```
khawkins, line 97: This...is not my favorite, relying on fatal errors in the shell to take affirmative alternative action. It can be risky, and is otherwise not super intuitive. A more standard and maintainable way to do this would be to create a basic Node utility module to manage the logic. We know we have a Node.js environment in either case, so we can write a pretty simple OS-agnostic script with `fs`, `path`, and `os` to do the copying, give it an identifiable name, and voila, maintainable code. 🙂

It's slightly more code than this, but IMO it's easier to maintain, and more intuitive to understand.

haifeng-li-at-salesforce (approved):
LGTM. Great job!

package.json
sfdctaka: Let me take care in another PR. (This PR got merged few minutes ago.)

I think I remember why I had `yaml` folder under `resources` folder. It's because I was able to use `cp -r`. Since the yaml file was move and `cp` was failing on copying a single file claiming it can't find `resources` folder.

src/lsp/client/client.ts
```diff
@@ -60,8 +60,8 @@ export function activate(
 
     // Create the language client and start the client.
     client = new LanguageClient(
-        'LSP Mobile',
-        'LSP Mobile Client',
+        'Salesforce Mobile Language Server',
```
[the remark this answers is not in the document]
sfdctaka: 👍 
Will take care of this along with the above feedback that you made.

PR #165, opened 2024-12-06
W-16620689 Add md files for [LWC Auth Debug] LSP to reduce OOM issues story

clivewong:


ben-zhang-at-salesforce (approved):
looks good to me

ben-zhang-at-salesforce:
fyi:  We don't need doc link for mobieOfflineFriendly.ts due to sf core vscode will render inline help when mouse over the component, in the help there's link pointing to external doc for the component. if the component is mobile friendly, there will be a tag there. 
image

PR #166, opened 2024-12-11
Md file link for diagnostic

ben-zhang-at-salesforce:
Clive finished the md file writeup in [this merged pr](https://github.com/salesforce/salesforcedx-vscode-mobile/pull/165).  here we updated the code to point to the md file when diagnostic is rendered over mouse over.   

It looks like below and now it matches what eslint plugin reported problems for vscode now :)  
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     },
     "repository": {
         "type": "git",
-        "url": "https://github.com/salesforce/salesforcedx-vscode-mobile.git"
+        "url": "https://github.com/salesforce/salesforcedx-vscode-mobile"
     },
     "license": "SEE LICENSE IN LICENSE.txt",
     "icon": "images/package-icon.png",
--- a/src/lsp/client/client.ts
+++ b/src/lsp/client/client.ts
@@ -22,6 +22,10 @@ export function activate(
     updateDiagnosticsSettingCommand: string,
     diagnosticsSettingSection: string
 ) {
+    // Get extension name
+    const extensionTitle =
+        context.extension.packageJSON.contributes.configuration.title;
+
     // The server is implemented in node
     const serverModule = context.asAbsolutePath(
         path.join('out', 'src', 'lsp', 'server', 'server.js')
@@ -54,7 +58,8 @@ export function activate(
         },
         initializationOptions: {
             updateDiagnosticsSettingCommand,
-            diagnosticsSettingSection
+            diagnosticsSettingSection,
+            extensionTitle
         }
     };
 
--- a/src/lsp/client/client.ts
+++ b/src/lsp/client/client.ts
@@ -7,6 +7,7 @@
 
 import * as path from 'path';
 import { workspace, ExtensionContext } from 'vscode';
+import { version, repository } from '../../../package.json';
 
 import {
     LanguageClient,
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     },
     "repository": {
         "type": "git",
-        "url": "https://github.com/salesforce/salesforcedx-vscode-mobile.git"
+        "url": "https://github.com/salesforce/salesforcedx-vscode-mobile"
```
ben-zhang-at-salesforce, line 24: don't really need the .git in the url, and then can use it for the full md file link generation

src/lsp/client/client.ts
```diff
@@ -22,6 +23,12 @@ export function activate(
     updateDiagnosticsSettingCommand: string,
     diagnosticsSettingSection: string
 ) {
+    // Get extension name
+    const extensionTitle =
+        context.extension.packageJSON.contributes.configuration.title;
+
+    const diagnosticBaseRootUrl = `${repository.url}/blob/v${version}/src/lsp/docs`;
```
ben-zhang-at-salesforce, line 30: get the lsp doc md file root url and pass to lsp server.

src/lsp/server/diagnostics/DiagnosticProducer.ts
```diff
@@ -8,19 +8,41 @@
 import { Diagnostic } from 'vscode-languageserver/node';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 
-export interface DiagnosticProducer<T> {
+export abstract class DiagnosticProducer<T> {
+    baseDocUrl: string;
```
haifeng-li-at-salesforce, line 12: Should it be declared as 'private' since by default, no accessor mean 'public'

package.json
khawkins: An alternative would be to use the [homepage](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#homepage) field of the configuration.

khawkins: Or maybe we just do it, since technically with GitHub repos you can clone the repo with `git clone` using the root URL (i.e. minus the `.git` extension), and it works. I guess as long as it works for your VCS provider, it's fine.

src/lsp/server/diagnostics/html/mobileOfflineFriendly.ts
```diff
@@ -17,11 +17,19 @@ const baseComponentsAttributes = { values: getBaseComponentsAttributes() };
...
+        super(baseDocUrl);
+    }
+
     getId(): string {
         return RULE_ID;
     }
 
+    hasDocumentation(): boolean {
+        return false;
+    }
```
haifeng-li-at-salesforce, line 31: note: MobileOfflineFriendly diagnostic does not has its own MD file. The tool tip for component from other SF tool has the offline indicator within the doc.

package.json
haifeng-li-at-salesforce: The `homepage`  is `https://github.com/salesforce/salesforcedx-vscode-mobile/blob/main/README.md`. If we change it to `https://github.com/salesforce/salesforcedx-vscode-mobile/`, will that change the UI content for mobile extension?

image

src/lsp/server/diagnostics/DiagnosticProducer.ts
ben-zhang-at-salesforce: done

package.json
khawkins: Possibly. Yeah I'd say just leave it. As long as the URL can be used to clone the repo (which it can), I'd say that covers our bases.

haifeng-li-at-salesforce (approved):
Code is well-organized and polished. Great job!

PR #174, opened 2025-03-25
feat(telemetry): send metrics to Azure 

sfdctaka:
### What does this PR do?
Executing onboarding wizard command and configure linting command will send telemetry info to Azure AppInsights.


package.json
```diff
@@ -41,7 +41,7 @@
         "menus": {
             "commandPalette": [
                 {
-                    "command": "salesforcedx-vscode-offline-app.configureLintingTools",
```
haifeng-li-at-salesforce, line 44: Why do we change command from camelCase into Kebab-case?

haifeng-li-at-salesforce (approved):
The improvement looks great!

package.json
sfdctaka: Good question. I cherry picked Angela's change for tooling hub because that contained code for handling telemetry. She was using kebab case there so I tried to match it. But then I yanked out her tooling hub code for the most part, except the telemetry handling code, because we aren't going to offer tooling hub feature anymore. If you want the camelCase we can revert it back after this PR gets merged in another PR.

haifeng-li-at-salesforce: As soon as it does not break the functionality, it is fine to keep it.

